PR #10, opened 2022-05-10
[MBUILDCACHE-18] Use m-resolver transport layer instead of redefining another one

gnodet:


pom.xml
```diff
@@ -142,6 +142,11 @@ under the License.
             <artifactId>maven-resolver-util</artifactId>
             <version>${resolverVersion}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.maven.resolver</groupId>
+            <artifactId>maven-resolver-transport-http</artifactId>
+            <version>${resolverVersion}</version>
```
cstamas, line 148: This one is in maven 3.9.x and 4.0.x... so maybe "provided"? Or build cache is meant to work with older versions as well?

src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
```diff
@@ -147,50 +157,33 @@ public void saveArtifactFile( CacheResult cacheResult,
...
-                LOGGER.info( "Cannot download {}, status code: {}", url, statusCode );
-                return Optional.empty();
-            }
-            try ( InputStream content = response.getEntity().getContent() )
-            {
-                return Optional.of( IOUtils.toByteArray( content ) );
-            }
+            GetTask task = new GetTask( new URI( url ) );
+            transporter.get( task );
+            return Optional.of( task.getDataBytes() );
```
cstamas, line 165: Word of warning: unsure the data size here, but if not "heap friendly" sized, this is OOM prone

cstamas (approved):
LGTM along with OOM remarks (I did not dig into build cache to check these spots)

pom.xml
gnodet: No, build cache requires maven >= 3.9.x, so I'll change it to provided.

jira-importer:
Resolve #248

jira-importer:
Resolve #248

PR #14, opened 2022-05-16
[MBUILDCACHE-20] Use local cache before remote cache by default

gnodet:
@maximilian-novikov @maximilian-novikov-db could you have a look at this one ?
I don't really understand why the remote cache would have to be checked before the local one...

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
+++ b/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
@@ -166,22 +166,22 @@ public class CacheControllerImpl implements CacheController
         ProjectsInputInfo inputInfo = projectInputCalculator.calculateInput( project );
 
         final CacheContext context = new CacheContext( project, inputInfo, session );
-        // remote build first
-        CacheResult result = findCachedBuild( mojoExecutions, context );
+        // local build first
+        CacheResult result = findLocalBuild( mojoExecutions, context );
 
         if ( !result.isSuccess() && result.getContext() != null )
         {
-            LOGGER.debug( "Remote cache is incomplete or missing, trying local build" );
+            LOGGER.debug( "Local cache is incomplete or missing, trying remote build" );
 
-            CacheResult localBuild = findLocalBuild( mojoExecutions, context );
+            CacheResult remoteBuild = findCachedBuild( mojoExecutions, context );
 
-            if ( localBuild.isSuccess() || ( localBuild.isPartialSuccess() && !result.isPartialSuccess() ) )
+            if ( remoteBuild.isSuccess() || ( remoteBuild.isPartialSuccess() && !result.isPartialSuccess() ) )
             {
-                result = localBuild;
+                result = remoteBuild;
             }
             else
             {
-                LOGGER.info( "Local build was not found by checksum " + inputInfo.getChecksum() );
+                LOGGER.info( "Remote build was not found by checksum " + inputInfo.getChecksum() );
             }
         }
         cacheResults.put( getVersionlessProjectKey( project ), result );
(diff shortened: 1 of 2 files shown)


maximilian-novikov-db:
> @maximilian-novikov @maximilian-novikov-db could you have a look at this one ? I don't really understand why the remote cache would have to be checked before the local one...

that's the good question, probably that's the legacy logic, which doesn't make sense after @AlexanderAshitkin added 'proper' plugins input reconciliation

i would just fix this, without introducing a new flag

gnodet:
> > @maximilian-novikov @maximilian-novikov-db could you have a look at this one ? I don't really understand why the remote cache would have to be checked before the local one...
> 
> that's the good question, probably that's the legacy logic, which doesn't make sense after @AlexanderAshitkin added 'proper' plugins input reconciliation
> 
> i would just fix this, without introducing a new flag

Ok, I'll simplify the PR then.

maximilian-novikov-db (approved):

AlexanderAshitkin:
> > > @maximilian-novikov @maximilian-novikov-db could you have a look at this one ? I don't really understand why the remote cache would have to be checked before the local one...
> 
> > 
> 
> > that's the good question, probably that's the legacy logic, which doesn't make sense after @AlexanderAshitkin added 'proper' plugins input reconciliation
> 
> > 
> 
> > i would just fix this, without introducing a new flag
> 
> 
> 
> Ok, I'll simplify the PR then.

Hi
The idea was to use the same artifact across all the cache consumers. By looking in the remote first it's possible to avoid using locally built artifacts and cached binaries should become eventually consistent across all nodes, even if remote cache was overridden. I cannot recall other implications except that, but discrepancies theoretically possible if binaries compared directly(snapshots?). But definitely local first lookup will work in majority of cases

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -166,22 +166,22 @@ public CacheResult findCachedBuild( MavenSession session, MavenProject project,
         ProjectsInputInfo inputInfo = projectInputCalculator.calculateInput( project );
 
         final CacheContext context = new CacheContext( project, inputInfo, session );
-        // remote build first
-        CacheResult result = findCachedBuild( mojoExecutions, context );
```
AlexanderAshitkin, line 170: @gnodet  @maximilian-novikov-db remote honestly this change brings regression.
There are 2 local caches: 
1) local cache of remote repository
2) cache of local builds

Previously, if remote build was present in local cache it was always used - there was no lookup. Now local build used with priority which could lead to difficult to debug issues.

gnodet: Well, that was exactly what I was trying to fix.  If there was no remote build cached, but a local build, a lookup did occur...

AlexanderAshitkin: > @AlexanderAshitkin Well, that was exactly what I was trying to fix. If there was no remote build cached, but a local build, a lookup did occur... This basically means that local builds are useless if you have a remote cache, because they are never used, which is a bit weird.

Not exactly. This is lookup for a local build (which is not from remote cache)
```
        // local build first
        CacheResult result = findLocalBuild( mojoExecutions, context );
```
This is lookup to find cached build from remote repo (if missed - then lookup remote cache):
```
    CacheResult remoteBuild = findCachedBuild( mojoExecutions, context );
```
Old logic was:
1) Lookup local cache of remote repo
2) If #1 failed lookup remote repo and download
3) If #2 failed (effectively no remote build) - rebuild locally

Now logic is:
1) Try to use local build even if build is present in remote repo
2) Try to find locally cached remote build
3) Loockup remote cache

The new scheme has minimal performance benefits (both tried to find local build first), but likely will lead to less consistent results and difficult to debug discrepancies in presence of remote cache

gnodet:
I reverted the change.

jira-importer:
Resolve #232

PR #33, opened 2022-10-30
[MBUILDCACHE-32] Do not print exception when probing builds

AlexanderAshitkin:
Part of the cache is to discover remote caches by sending GET http requests. These requests are exploratory probes in nature and assume 404 Not Found errors by design. Such errors should not be printed in logs with exceptions, because that makes logs look like there are errors

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed 
       pulling in other changes.
 - [x] Format the pull request title like `[MNG-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
+++ b/src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
@@ -111,7 +111,7 @@ public class RemoteCacheRepositoryImpl implements RemoteCacheRepository, Closeab
     }
 
     @Override
-    public boolean getArtifactContent(CacheContext context, Artifact artifact, Path target) throws IOException {
+    public boolean getArtifactContent(CacheContext context, Artifact artifact, Path target) {
         return getResourceContent(getResourceUrl(context, artifact.getFileName()), target);
     }
 
@@ -145,19 +145,32 @@ public class RemoteCacheRepositoryImpl implements RemoteCacheRepository, Closeab
      * @return null or content
      */
     @Nonnull
-    public Optional<byte[]> getResourceContent(String url) throws IOException {
+    public Optional<byte[]> getResourceContent(String url) {
+        String fullUrl = getFullUrl(url);
         try {
-            LOGGER.info("Downloading {}", getFullUrl(url));
+            LOGGER.info("Downloading {}", fullUrl);
             GetTask task = new GetTask(new URI(url));
             transporter.get(task);
             return Optional.of(task.getDataBytes());
-        } catch (Exception e) {
-            LOGGER.info("Cannot download {}", getFullUrl(url), e);
+        } catch (ResourceDoesNotExistException e) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Cache item not found: {}", fullUrl, e);
+            } else {
+                LOGGER.info("Cache item not found: {}", fullUrl);
+            }
             return Optional.empty();
+        } catch (Exception e) {
+            if (cacheConfig.isFailFast()) {
+                LOGGER.error("Error downloading cache item: {}", fullUrl, e);
+                throw new RuntimeException("Error downloading cache item: " + fullUrl, e);
+            } else {
+                LOGGER.error("Error downloading cache item: {}", fullUrl);
+                return Optional.empty();
+            }
         }
     }
 
-    public boolean getResourceContent(String url, Path target) throws IOException {
+    public boolean getResourceContent(String url, Path target) {
         try {
             LOGGER.info("Downloading {}", getFullUrl(url));
             GetTask task = new GetTask(new URI(url)).setDataFile(target.toFile());
--- a/src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
+++ b/src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
@@ -25,6 +25,7 @@ import javax.inject.Named;
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -32,6 +33,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.http.HttpStatus;
+import org.apache.http.client.HttpResponseException;
 import org.apache.maven.SessionScoped;
 import org.apache.maven.buildcache.checksum.MavenProjectInput;
 import org.apache.maven.buildcache.xml.Build;
@@ -153,13 +156,17 @@ public class RemoteCacheRepositoryImpl implements RemoteCacheRepository, Closeab
             transporter.get(task);
             return Optional.of(task.getDataBytes());
         } catch (ResourceDoesNotExistException e) {
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.info("Cache item not found: {}", fullUrl, e);
-            } else {
-                LOGGER.info("Cache item not found: {}", fullUrl);
-            }
+            logNotFound(fullUrl, e);
             return Optional.empty();
         } catch (Exception e) {
+            // this can be wagon used so the exception may be different
+            // we want wagon users not flooded with logs when not found
+            if ((e instanceof HttpResponseException
+                            || e.getClass().getName().equals(HttpResponseException.class.getName()))
+                    && getStatusCode(e) == HttpStatus.SC_NOT_FOUND) {
+                logNotFound(fullUrl, e);
+                return Optional.empty();
+            }
             if (cacheConfig.isFailFast()) {
                 LOGGER.error("Error downloading cache item: {}", fullUrl, e);
                 throw new RuntimeException("Error downloading cache item: " + fullUrl, e);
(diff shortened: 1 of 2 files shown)


maximilian-novikov-db (approved):

kwin:
There are lots of unrelated documentation fixes. Can you separate those into a dedicated PR?

AlexanderAshitkin:
> There are lots of unrelated documentation fixes. Can you separate those into a dedicated PR?

I would prefer not to do so. The change itself is trivial, the documentation changes are also ok as I understand. Please approve if you don't have objections on the essence of changes.

kwin:
Both should end up in separate commits, otherwise finding git commits introducing regressions is much harder than it should be.

AlexanderAshitkin:
> Both should end up in separate commits, otherwise finding git commits introducing regressions is much harder than it should be.

Regression will not happen because of the documentation changes. Code change is isolated and fixes the very clear issue. I'm happy to address any issues with documentation and the code. Let's not overcomplicate this simple change.

src/site/markdown/remote-cache.md
```diff
@@ -88,13 +99,12 @@ Allow writes in remote cache add jvm property to designated CI builds.
...
-### Remote cache relocation to local builds
+### Make remote cache portable
 
-As practice shows, developers often don't realize that builds they run in local and CI environments are different. So
-straightforward attempt to reuse remote cache in local build usually results in cache misses because of difference in
-plugins, parameters, profiles, environment, etc. In order to reuse results you might need to change poms, cache config,
-CI jobs and the project itself. This part is usually most challenging and time-consuming. Follow steps below to
-iteratively achieve working configuration.
+As practice shows, builds which run on local workstations and CI environments are different. Straightforward attempt to
+reuse cache produced in different environment usually results in cache misses. As is, caches are rarely portable because
```
gnodet, line 105: `reuse the cache`
`in a different`

src/site/markdown/remote-cache.md
```diff
@@ -112,7 +122,7 @@ Copy the link to a `build-cache-report.xml` and provide it to your local build a
 mvn verify -Dmaven.build.cache.failFast=true -Dmaven.build.cache.baselineUrl=https://your-cache-url/<...>/915296a3-4596-4eb5-bf37-f6e13ebe087e/build-cache-report.xml
 ```
 
-Once discrepancy between remote and local builds detected cache will fail with diagnostic info in
+Once discrepancy between remote and local builds detected cache will fail and spool diagnostic info in
```
gnodet, line 125: `Once the discrepancy between remote and local versions is detected, the cache will fail and diagnostic information will be stored in`

src/site/markdown/remote-cache.md
```diff
@@ -123,7 +133,7 @@ project's `target/incremental-maven` directory:
 
 Review `buildsdiff.xml` file and eliminate detected discrepancies. You can also diff build-info files directly to get
 low level insights. See techniques to configure cache in [How-To](how-to.md) and troubleshooting of typical issues
-in the section below.
+in the section below. Also, it is possible to diff `buildInfo.xml` from remote cache and local workstation directly.
```
gnodet, line 136: `from` -> `between`

gnodet (changes requested):

src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
```diff
@@ -165,9 +166,28 @@ public Optional<byte[]> getResourceContent( String url ) throws IOException
...
         {
-            LOGGER.info( "Cannot download {}", getFullUrl( url ), e );
+            if ( LOGGER.isDebugEnabled() )
+            {
+                LOGGER.debug( "Cannot download cache item {}", getFullUrl( url ), e );
+            }
+            else
+            {
+                LOGGER.error( "Cannot download cache item {}: {}", getFullUrl( url ), e );
+            }
```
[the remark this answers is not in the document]
olamy: > Can I suggest that if we get here it should be at least a warning regardless of whether debug is enabled. I'd go for error() as the remote cache can be easily disabled if it is known that it's not available.

not sure to understand. if the remote cache is disabled. no download will be tried.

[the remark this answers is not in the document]
alexthomasmc: Yes that's the scenario I had in mind, i.e. enabling the remote cache is a deliberate act so if it then doesn't work that should be an error not a warning.
In the situation where the user doesn't expect the remote cache to work, e.g. working offline, they should disable it.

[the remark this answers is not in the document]
PayBas: > i.e. enabling the remote cache is a deliberate act so if it then doesn't work that should be an error not a warning.
In the situation where the user doesn't expect the remote cache to work, e.g. working offline, they should disable it.

I don't think that's the intended scenario described by the author.

The remote cache can be working 100% correctly and still not contain the requested items. In fact, this is 100% bound to happen, unless the cache is magically warmed-up beforehand.

1. New project/branch/version/whatever so cache is empty.
2. Build starts
3. Cache extension cannot find cache item of Module X. <-- this is 100% normal and expected behavior and shouldn't throw an error (or even a warming i.m.h.o.), let alone a stack-trace.
4. Module X gets built
5. Module X output gets put into the cache
6. Next build starts, and now it can find cache item of Module X.

Having the logs filled with error/stack-traces for non-issues is counterproductive. So I support any attempt at a more sane output.

src/test/java/org/apache/maven/buildcache/its/RemoteCacheDavTest.java
```diff
@@ -68,7 +68,7 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @IntegrationTest("src/test/projects/remote-cache-dav")
-@Testcontainers(disabledWithoutDocker = true)
+@Testcontainers
```
olamy, line 71: why?
at least tests doesn't fail if docker is not available

olamy (changes requested):

src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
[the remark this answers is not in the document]
ferdnyc: @alexthomasmc Is this the remote cache "not working", though, or can this also just represent a cache miss / expired cache?

Cache misses certainly are not errors, heck they're not even problems, they're a standard aspect of how caching works. Data stored in a cache can be wiped out at any moment, for any reason (or even no reason at all). Caches go missing all the time.

When a build system fails to obtain a cached artifact, it simply needs to generate it anew. It can do that, it has the technology.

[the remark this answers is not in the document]
alexthomasmc: The cache miss exception is caught just before this, so by the time we get here there's a bigger problem. Agree with your policy on this - I think that's the intent of the current logic, it's just the logging that's the issue.

src/main/java/org/apache/maven/buildcache/RemoteCacheRepositoryImpl.java
```diff
@@ -144,19 +145,32 @@ public void saveArtifactFile(CacheResult cacheResult, org.apache.maven.artifact.
...
-            LOGGER.info("Downloading {}", getFullUrl(url));
+            LOGGER.info("Downloading {}", fullUrl);
             GetTask task = new GetTask(new URI(url));
             transporter.get(task);
             return Optional.of(task.getDataBytes());
-        } catch (Exception e) {
-            LOGGER.info("Cannot download {}", getFullUrl(url), e);
+        } catch (ResourceDoesNotExistException e) {
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Cache item not found: {}", fullUrl, e);
```
gnodet, line 157: I think it should be `LOGGER.info`

olamy: @gnodet there is `if (LOGGER.isDebugEnabled())` and else `info` and, more important without the exception as the last parameter, which is very verbose. per default, we will be fine. 
But could be really more simple to remove this if and using only ` LOGGER.info("Cache item not found: {}", fullUrl);`

alexthomasmc: Maybe? It's a tradeoff - you want to know how well the cache is working but you don't want to clutter the log (in some of our builds we have nearly 100 modules). Also the "not found" phrase implies to me that something could be wrong but of course a cache miss is completely normal, so perhaps "Item not in cache" would sound more innocent. 
A summary at the end of how much the cache was used instead would be very nice...

gnodet: @olamy I'm fine with displaying the exception if debug is enabled, but the level of the message is part of the semantic and should not really change depending on which level is activated.
I've used this pattern a log, but I think it should be:
```
if (isDebugEnabled()) {
   log.info(xxx, exception);
} else {
   log.info(xxx);
}
```

olamy: @gnodet, why not I'm fine with that. it just looks weird to me :)

olamy: > Maybe? It's a tradeoff - you want to know how well the cache is working but you don't want to clutter the log (in some of our builds we have nearly 100 modules). Also the "not found" phrase implies to me that something could be wrong but of course a cache miss is completely normal, so perhaps "Item not in cache" would sound more innocent. A summary at the end of how much the cache was used instead would be very nice...

agree I'm using the cache for a build with 328 modules so the current log is very very verbose when the cache not found :)

gnodet: Then the individual statements could be at debug level (with exception at trace level eventually), while the summary could be logged at info level.

olamy: go with that. But as we are in the case of `ResourceDoesNotExistException` not sure why do we need to have a `if` and display the exception. 
no if but just:
```
} catch (ResourceDoesNotExistException e) {
  LOGGER.info("Cache item not found: {}", fullUrl);
}
```

We are just in the case of a resource not found so no need to print the exception the cause is really clear. I'm not sure to understand of printing a stack even in debug mode. Is there anything more interested in the stack?

Frankly I don't even understand why an exception is used for that. There should be simply a boolean `not found` and not wasting CPU/memory by using an exception (but I know it's a different problem :))

olamy: could we make progress here? It would be fine to have this done and released as it is very verbose for no gain

olamy: @AlexanderAshitkin ping :)

jira-importer:
Resolve #257

PR #34, opened 2022-10-31
[MBUILDCACHE-25] Fixes N+1 cache processings for forked executions

AlexanderAshitkin:
Forked execution should be thought as a part of originating mojo internal implementation. If forkedExecution is detected, it means that originating mojo is not cached so fork should rerun too. If originating mojo is cached, it is cached altogether, with forks. The fix detects forked execution and disables cache processing if forked project entered cache.

Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 - [x] Format the pull request title like `[MNG-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MNG-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
       be performed on your pull request automatically.
 - [x] You have run the [Core IT][core-its] successfully.

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under 
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

[core-its]: https://maven.apache.org/core-its/core-it-suite/

PR #43, opened 2023-02-16
Add maven.build.cache.remote.enabled parameter

atsteffen:
Use-case:
You may have cases where you want to setup a remote cache, but for whatever reason developers or some build machines do not have access to the server, or have slow internet connection that makes downloading artifacts not sufficiently advantageous over building and caching locally only.

With this parameter, the remote cache can be disabled by default and selectively enabled using a -Dmaven.build.cache.remote.enabled=true flag.

Note:
I still need to get a ASF JIRA account in order to make an issue for this request.  I've joined the maven mailing list, but I'm still unclear who to email to request this. There are some issues I'd like to investigate further and potentially contribute PR for on this project.

 - [ ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed 
       pulling in other changes.
 - [ ] Format the pull request title like `[MNG-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [ ] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 

 - [ x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


olamy:
created here https://issues.apache.org/jira/browse/MBUILDCACHE-46

olamy:
if remote cache is enabled/disabled via configuration does this correctly override it? 
@gnodet  I wonder in configuring remote url via cli would be useful as well?

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
```diff
@@ -452,7 +453,9 @@ public boolean isEnabled() {
     @Override
     public boolean isRemoteCacheEnabled() {
         checkInitializedState();
-        return getRemote().getUrl() != null && getRemote().isEnabled();
+        return getRemote().getUrl() != null
+                && Boolean.parseBoolean(System.getProperty(
```
michael-o, line 457: No, use properties from the session.

michael-o (changes requested):

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
atsteffen: Thanks!  I went ahead and updated a couple other Boolean.getProperty(..) methods as those also go to the system properties.

olamy (approved):

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
```diff
@@ -597,4 +597,15 @@ private String getProperty(String key, String defaultValue) {
...
+
+    private boolean getProperty(String key, boolean defaultValue) {
+        String value = session.getUserProperties().getProperty(key);
+        if (value == null) {
+            value = session.getSystemProperties().getProperty(key);
+            if (value == null) {
+                return defaultValue;
+            }
+        }
+        return Boolean.parseBoolean(value);
```
michael-o, line 609: I think with system properties this is redundant, no?

AlexanderAshitkin:
looks good to me

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -173,11 +173,14 @@ public CacheResult findCachedBuild(
             LOGGER.info("Attempting to restore project {} from build cache", projectName);
 
             // remote build first
-            result = findCachedBuild(mojoExecutions, context);
+            if (cacheConfig.isRemoteCacheEnabled()) {
```
AlexanderAshitkin, line 176: Please write integration tests for this flag. In pr #34, I added Wiremock-based integration tests to verify the remote cache behavior. The same approach could be reused here. It will be good to see a test that checks that remote is disabled.

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
```diff
@@ -452,28 +453,29 @@ public boolean isEnabled() {
     @Override
     public boolean isRemoteCacheEnabled() {
         checkInitializedState();
-        return getRemote().getUrl() != null && getRemote().isEnabled();
```
AlexanderAshitkin, line 455: Though nothing changed here, the `RemoteCacheEnabled` flag intuitively should disable all the remote interactions. So all the flags below should be preconditioned on the `remote. enabled`. Again, there is no regression in the current form, and this is an optional consideration.

AlexanderAshitkin (changes requested):
It will be good to:
1) Write an integration test because flag combinations could result in a not-so-obvious behavior.
2) [optionally] Precondition all the other remote flags on the `remote.enabled`.

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
atsteffen: This could be accomplished any number of ways, but I decided to go with the micro-optimized form that does not require a string parse to handle the default case.

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
atsteffen: 1. For testing flag and xml setting combinations, a lightweight unit test seems prudent.  I went ahead and added a unit test for CacheConfigImpl.  I tested the initialization and the remote parameters only, but structured it in a way that it should easily accommodate tests around other configurations.  A couple integration tests make sense as well, but I thought i'd wait until PR #34 is merged to build off that approach. Tests are a great way to guide meaningful cleanups and refactors, and I feel I need some more time and context to do this at an integration level.

2. Remote configurations are now preconditioned on any meaningful encompassing configuration behind the CacheConfig interface.

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
atsteffen: Added

AlexanderAshitkin (approved):

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
AlexanderAshitkin: Thank you, I clicked it!

michael-o (approved):
As far as my concerns have been addressed, I am fine with this change.

jira-importer:
Resolve #312

PR #51, opened 2023-03-21
[MBUILDCACHE-48] - Add possibility to disable caching per module in a multi-module project

eltsovalex:
Added a new flag that allows disabling cache (not just looking up build artifacts, but all the build cache completely).
This may be required in a case when a module is a final leaf of a reactor tree, and its artefacts are not required by other modules as cached dependencies. Hence eliminating some modules from caching will speed up build and also lower the footprint size of a cached build.

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MNG) filed 
       pulling in other changes.
 - [x] Format the pull request title like `[MNG-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
+++ b/src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
@@ -99,12 +99,18 @@ public class BuildCacheMojosExecutionStrategy implements MojosExecutionStrategy 
         CacheState cacheState = DISABLED;
         CacheResult result = CacheResult.empty();
         boolean skipCache = cacheConfig.isSkipCache() || MavenProjectInput.isSkipCache(project);
+        boolean cacheIsDisabled = MavenProjectInput.isCacheDisabled(project);
         if (source == Source.LIFECYCLE) {
             List<MojoExecution> cleanPhase = lifecyclePhasesHelper.getCleanSegment(project, mojoExecutions);
             for (MojoExecution mojoExecution : cleanPhase) {
                 mojoExecutionRunner.run(mojoExecution);
             }
-            cacheState = cacheConfig.initialize();
+            if (!cacheIsDisabled) {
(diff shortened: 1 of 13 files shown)


AlexanderAshitkin:
Hi Alex. 
Please clarify the use case. Is it for cases where building from scratch is faster than using the cache? I understand uploading might be expensive, but an existing configuration parameter allows for skipping the upload. And if uploading doesn't happen, I wonder what the benefits are achieved by disabling the cache compared to disabled upload.
Thank you!

eltsovalex:
> Hi Alex. Regarding the code itself - looks good to me. But please clarify the use case and rationale for the change. Is it for cases where the local build is faster than downloading from the cache? I understand uploading might be expensive, but an existing configuration parameter allows for skipping the upload. And if uploading doesn't happen, I wonder what the benefits are achieved by disabling the cache compared to disabled upload. Thank you!

Hi. Thanks for asking. Sorry for not providing a precise description of my use-case.
So, I have a module in my project that builds a huge EAR of ~200Mb as a target. Due to some specific reasons (e.g. project version written deep down in one of jars inside EAR +some build artifacts required for publishing by CI to be present in target folder) I have to always build that subproject clean (so without caching). And while the previous flag I added last year (maven.build.cache.skipCache) works fine to solve the forced rebuild task - it still uploads the 200Mb EAR to cache. So it spoils the cache with multiple versions of a huge unused file + takes some time for the upload itself. As this is not an attached artefact, but a main goal - I can not add it to ignored/not uploaded.  
Hence this new property added that totally disables cache on a per-project level (both lookup and upload)

maximilian-novikov-db (approved):

AlexanderAshitkin (approved):

olamy:
worth to document the behaviour https://maven.apache.org/extensions/maven-build-cache-extension/parameters.html
e.g this parameter can apply to single module(s) as well.

jira-importer:
Resolve #307

PR #86, opened 2023-06-06
[MBUILDCACHE-60] Remove undeclared Guava dependency with Arrays.asList

elharo:


elharo:
ping

olamy:
> ping

why ping?
if CI is green just merge it. no need to ask review for such a minor change because it's just modifying test sources not sure you need to ping people to validate this.

elharo:
I can't agree with that. Code review is essential. If it's that trivial, then it's an easy review. If it isn't, it needs review all the more. I've certainly noticed more than one mistake in Maven when code review was skipped, both by myself and others.

src/test/java/org/apache/maven/buildcache/LifecyclePhasesHelperTest.java
```diff
@@ -19,9 +19,9 @@
 package org.apache.maven.buildcache;
 
 import java.util.ArrayList;
```
ferdnyc, line 21: I notice you left the `ArrayList` import here (as opposed to in the other files), is that still needed?

ferdnyc (commented):
Looks sensible enough to me.

ferdnyc:
I'm with @elharo — I _hate_ merging my own code to a public repo, and will generally only do it if it's a non-code change or I can definitively state that merging won't impact other users.

Otherwise, I figure a second pair of eyes is a sensible precaution on almost **any** code change, and when there are multiple people with commit access a good general rule is to have someone other than the PR author do the merge. I try to review others' changes as much as I can, in the hopes they'll return the favor.

In that spirit, I've pitched in with an inexpert review of the changes here.

src/test/java/org/apache/maven/buildcache/LifecyclePhasesHelperTest.java
elharo: Yes, it's used in a couple of unchanged lines in this class

gnodet (approved):

jira-importer:
Resolve #331

jira-importer:
Resolve #331

PR #92, opened 2023-08-03
[MBUILDCACHE-67] Bugfix artefact restoration error not handled properly

kbuntrock:
### Description of this pull request

https://issues.apache.org/jira/projects/MBUILDCACHE/issues/MBUILDCACHE-67

- Fix artefact restoration error was not executing the regular project build as expected
- Fix non-lazy artefact restoration was in fact partially lazy
- Added an IT test on this

Draft until PR-MBUILDCACHE-64 is merged since I want to factorize a function in the IT test (see the related todo).

### MR checklist

 - [x] Make sure there is a [MBUILDCACHE JIRA issue](https://issues.apache.org/jira/browse/MBUILDCACHE) filed 
       pulling in other changes.
 - [x] Format the pull request title like `[MBUILDCACHE-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.


src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -44,17 +44,15 @@
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.Future;
-import java.util.concurrent.FutureTask;
+import java.util.concurrent.*;
```
olamy, line 47: we don't like much import `*` :)

kbuntrock: Noted, will fix it. :)

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
```diff
@@ -129,7 +129,8 @@ public void execute(
             boolean restored = result.isSuccess(); // if partially restored need to save increment
             if (restorable) {
                 restored &= restoreProject(result, mojoExecutions, mojoExecutionRunner, cacheConfig);
-            } else {
+            }
+            if (!restored) {
```
maximilian-novikov-db, line 133: perhaps we need to execute all mojos in this case including for Clean phase to remove any "garbage" which could left after an unsuccessful restore

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -401,6 +399,26 @@ private Future<File> createDownloadTask(
         });
         if (!cacheConfig.isLazyRestore()) {
```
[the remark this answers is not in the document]
kbuntrock: "LazyRestore" seems fine to me. I would instead replace "download" by "restore" since there is no downloading involved when working with a local cache. But it's not something I will do in this PR. :P

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -401,6 +399,26 @@ private Future<File> createDownloadTask(
...
             downloadTask.run();
+            try {
+                downloadTask.get();
```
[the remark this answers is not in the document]
kbuntrock: As I apprehend the "lazy restore" functionality, there are two approach, each one with their pros and cons.

**Non lazy restore** (the default) : immediate restoration of artefact and other build outputs.
Build stability is the main concern. If the restoration fails, the build is executed normally. (I'm interrogative about the part where you say it can fail the build. For me it's the exact opposite intend). **Con**: we might restore stuff not used further in another module build.

**Lazy restore** : restoration of artefacts is done only when used. It can save some bandwidth and speedup the build when used with a remote cache. (Pros with a local cache are not so clear to me.) **Con**: If the cache is corrupted/not accessible and an artefact must be used as a dependency in another module : the build is doomed and will fail.

Actually, the non lazy restore does not really work as expected since the restore is done (download is immediately launched when using a remote cache), but we don't check the result straight away. It might be far too late when we'll discover the artefact did not restored well.

Am I missing something in the picture?

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
kbuntrock: Good idea @maximilian-novikov-db.

@AlexanderAshitkin : do you mean we should change the way we restore artefacts from remote caches? I'm more inclined to the "second clean phase" idea since it covers any restoration problem.

AlexanderAshitkin: That is discussable:
* `clean` phase has been completed by this time, and repeating it sounds slightly off to me. 
* Keeping partially restored artifacts is error-prone, and `clean` might not even be requested. 
* invoking ad-hoc `clean` might wipe out other cached data and might be undesirable

It is better to leave it as is and rely on the regular `clean`, or implement `mv` logic to minimize the risk of corrupted cache data in the target dir.

maximilian-novikov-db: > * Keeping partially restored artifacts is error-prone, and `clean` might not even be requested.

all mojos except clean are executed again if it falls back to the "normal" build, i'd say keeping any partially restored data is harmful and may lead to unpredictable result

AlexanderAshitkin: I meant that the `clean` might not be requested by the user. In that case, using the `clean` looks undesirable because it has side effects outside the cache that might confuse the user. And I agree that leaving partially restored data is error-prone. But all three options (leaving as is, adding clean, or moving) have drawbacks.

maximilian-novikov-db: i suggest to remove the filter here https://github.com/apache/maven-build-cache-extension/blob/4d53a0a05c1e006265e19bce8ba8c9a0eb7831df/src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java#L137

and not to add Clean if it's not requested by user

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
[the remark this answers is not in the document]
maximilian-novikov-db: @kbuntrock could you please move this change to a separate pr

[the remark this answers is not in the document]
kbuntrock: @maximilian-novikov-db  : I would rather not since I consider this as part of the bug I am willing to fix.

Here are some other arguments.

The documentation state the following in the "performance" page : 

> Use a lazy restore
> 
> By default, the cache tries to restore all artifacts for a project preemptively. Lazy restore could give significant time by avoiding requesting and downloading unnecessary artifacts from the cache. It is beneficial when small changes are a dominating build pattern. Use command line flag:
> 
>     -Dmaven.build.cache.lazyRestore=true";
> 
> In cache corruption situations, the lazy cache cannot support fallback to normal execution. It will fail instead. To heal the corrupted cache, manually remove corrupted cache entries or force cache rewrite.
 
Implied by this piece of documentation : Contrary to the lazy restore, the pre-emptive default restore option supports fallback. This is actually not the case for artifacts, the most important build outputs. 
Alongside, restoration of generated source and other attached outputs works as expected : any cache corruption will trigger a regular build.
I have a strong feeling we are just in presence of a simple implementation oversight I'm fixing here : calling "run" on a FutureTask does the restoration but do no throw any exception upon failure. It has to be checked via a get method.

@AlexanderAshitkin 
To complete a bit more this question : 
> Does it solve any practical issues? 

Yes. Actually, if a corrupted/non downloadable artefact is used in a dependant module while we have not used the "lazy restore" functionality : the build will fail.
Worst case before this fix : the build fails
Worst case after this fix : we do a regular build.

Seems pretty important that one should be able to choose between stability over rapidity.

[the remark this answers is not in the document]
AlexanderAshitkin: Agree. Thanks for the clarifications!

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -401,6 +399,26 @@ private Future<File> createDownloadTask(
         });
         if (!cacheConfig.isLazyRestore()) {
             downloadTask.run();
```
AlexanderAshitkin, line 401: run could be removed if we call `get`

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
kbuntrock: I agree that seeing a second clean phase seems a bit "off" but I could be in piece with it since it is a abnormal edge case (and we will write a beautiful log to explain it. :p)

Trying to summarize the exchange in order to check if we have a consensus in a situation where the cache fails the restoration
- the build was requested without clean phase : we do not change the actual behaviour.
- the build was requested with a clean phase : an extra clean phase will be executed before the actual behaviour (= the whole build is executed).

We change generically the log "Cannot restore project artifacts, continuing with non cached build" by : 
"Cannot restore project artifacts, continuing with non cached build (clean phase will be re-executed if requested)".

And in another PR, the remote download could be enhanced by doing it in two steps : downloading in a temp directory + moving it
Reducing the risk of leftovers files.

AlexanderAshitkin: Some clarifications about the current implementation:
* Logic of rerunning on failures is already within the BuildCacheMojosExecutionStrategy#restoreProject:
```java
if (!restored) {
    // cleanup partial state
    project.getArtifact().setFile(null);
    project.getArtifact().setResolved(false);
    mojoListener.remove(project);
    // build as usual
    for (MojoExecution mojoExecution : cachedSegment) { <<< rebuilding everything
        mojoExecutionRunner.run(mojoExecution);
    }
}
```
* `!restored` condition should be read as "at least part of the project were rebuilt, not restored"
* It doesn't mean an error automatically, it means something was rebuilt (at least in the design). 100% successful completion of the `restoreProject` could return `false`. 

So, in all, it feels like we might need fixes within the `restoreProject`:
1) to implement proper clean phase
2) Add a verification step to the "forced" steps - to force cache update on the detected inconsistencies
3) Make sure that in case of additional executions ("post-cached" or similar), we return `false` to update the cache.

kbuntrock: Alright, finally had some time to work on what you pointed out : regular build after a failed try to use the cache was handled differently depending the situation.
I will add comments on some lines of this commit to improve your ability to review it effectively. But here are some global points : 
- I moved the part where we check cached mojos consistency before artifact restoration. There are two benefits : this steps is faster than restoration over the network. So if we can't use the cache because of mojos parameters, better to know it before downloading artifacts. And secondly: there are no more "things" to rollback since setting artifacts of the project is done at the very end of the restoration.
- I kept the regular build execution outside the "restoreProject" function.
- I execute an extra clean phase only if two conditions are met : clean phase was requested in the build + there are potentially files restored in the project directory (it does not happen if we don't restore generated resources, "outputArtifacts" or if an error occurs before)  

I'm removing the draft on this PR since I don't want anymore to link it with my other PR. The duplicated 3 line function used in IT test can be remove in a very small 3rd PR.

src/main/java/org/apache/maven/buildcache/xml/CacheState.java
```diff
@@ -22,7 +22,6 @@
  * CacheState
  */
 public enum CacheState {
-    NOT_INITIALIZED,
```
kbuntrock, line 25: This enum value was not used.

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
kbuntrock: done

src/main/java/org/apache/maven/buildcache/MojoParametersListener.java
```diff
@@ -77,8 +77,4 @@ public void afterExecutionFailure(MojoExecutionEvent event) {
     public Map<String, MojoExecutionEvent> getProjectExecutions(MavenProject project) {
         return projectExecutions.get(project);
     }
-
-    public void remove(MavenProject project) {
```
kbuntrock, line 81: If I understood well, this was used to "reset" the project mojo's execution spying after a "rollback" from using the cache to a regular build execution.
I'm not anymore in a position where I need this function so I deleted it. But I could use your careful reviewing on it @maximilian-novikov-db (I see you name on the oldest commit of the class :p )

AlexanderAshitkin (approved):
It looks good to me. I added a few comments for potential optimizations, but those enhancements are out of the scope.

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
```diff
@@ -206,31 +259,20 @@ private boolean restoreProject(
...
+                            "Skipping plugin execution (cached): {}",
+                            cacheCandidate.getMojoDescriptor().getFullGoalName());
                 }
             }
 
+            // Execute mojos after the cache segment
+            LOGGER.debug("Execute mojos post cache segment");
             List<MojoExecution> postCachedSegment =
                     lifecyclePhasesHelper.getPostCachedSegment(project, mojoExecutions, build);
             for (MojoExecution mojoExecution : postCachedSegment) {
```
AlexanderAshitkin, line 272: @kbuntrock it is up to you, but please notice 2 edge cases here:
* A post-cached segment is possible if initially the build was cached at, say `package` phase and the current build runs a later phase, say `install`. In that case, the best behavior is to update the cached build with the more complete version, which won't happen in the `SUCCESS` case - maybe need a separate status.
* The same for forced executions - even if the execution is forced, it could be consistent or not. If it is not consistent, it could be desirable to update the cache. We shouldn't update the cache on discrepancies only if it is prohibited by configs.

kbuntrock:
Hello there,

A little up on this old-timer PR. If I'm correct I fixed every mandatory request. 

Any other thought?

maximilian-novikov-db (approved):

cowwoc:
I think I found a possible regression of this issue. Please take a look at https://issues.apache.org/jira/browse/MNG-8546

jira-importer:
Resolve #330

PR #98, opened 2023-08-24
Avoid hash calculation when <maven.build.cache.skipCache> is true 

atsteffen:
I have a large multi-module project with a top level project whose only responsibility is to wrap together the other modules along with some large resources into a deployable zip or docker image.  The top level project pulls together enough stuff that it gives an array size limit error when trying to calculate the cache for all of its inputs.  Since this top level project should always be rebuilt if any code changes, it is a good candidate for using <maven.build.cache.skipCache>true<maven.build.cache.skipCache>.  However, I found that using this parameter does not actually prevent the calculation of the hash.  It seems reasonable to consider this parameter as disabling all aspects of caching, including hash calculation and saving of the cached artifact.

I realize this change is rather surgical, and a more involved refactor might be warranted especially with respect to delicate boolean logic, and nullability of inputInfo in CacheContext.  However, before I start any refactoring I wanted to make sure this issue was filed and discussed.

 - [ ] Make sure there is a [MBUILDCACHE JIRA issue](https://issues.apache.org/jira/browse/MBUILDCACHE) filed 
       pulling in other changes.
 - [ ] Format the pull request title like `[MBUILDCACHE-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [ ] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 

 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
+++ b/src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
@@ -139,7 +139,7 @@ public class BuildCacheMojosExecutionStrategy implements MojosExecutionStrategy 
                 }
             }
 
-            if (cacheState == INITIALIZED && (!restorable || !restored)) {
+            if (cacheState == INITIALIZED && (!restorable || !restored) && !skipCache) {
                 final Map<String, MojoExecutionEvent> executionEvents = mojoListener.getProjectExecutions(project);
                 cacheController.save(result, mojoExecutions, executionEvents);
             }
--- a/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
+++ b/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
@@ -166,38 +166,35 @@ public class CacheControllerImpl implements CacheController {
 
         String projectName = getVersionlessProjectKey(project);
 
-        ProjectsInputInfo inputInfo = projectInputCalculator.calculateInput(project);
-
-        final CacheContext context = new CacheContext(project, inputInfo, session);
-
-        CacheResult result = empty(context);
-        if (!skipCache) {
-
-            LOGGER.info("Attempting to restore project {} from build cache", projectName);
-
-            // remote build first
-            if (cacheConfig.isRemoteCacheEnabled()) {
-                result = findCachedBuild(mojoExecutions, context);
-                if (!result.isSuccess() && result.getContext() != null) {
-                    LOGGER.info("Remote cache is incomplete or missing, trying local build for {}", projectName);
-                }
-            }
-
-            if (!result.isSuccess() && result.getContext() != null) {
-                CacheResult localBuild = findLocalBuild(mojoExecutions, context);
-                if (localBuild.isSuccess() || (localBuild.isPartialSuccess() && !result.isPartialSuccess())) {
-                    result = localBuild;
-                } else {
-                    LOGGER.info(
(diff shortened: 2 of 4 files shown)


src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -166,38 +166,35 @@ public CacheResult findCachedBuild(
 
         String projectName = getVersionlessProjectKey(project);
 
-        ProjectsInputInfo inputInfo = projectInputCalculator.calculateInput(project);
+        if (skipCache) {
+            LOGGER.info(
+                    "Project {} is marked as requiring force rebuild, will skip lookup in build cache", projectName);
```
elharo, line 171: delete log message as non-actionable

src/main/java/org/apache/maven/buildcache/DefaultProjectInputCalculator.java
```diff
@@ -72,7 +72,7 @@ public DefaultProjectInputCalculator(
 
     @Override
     public ProjectsInputInfo calculateInput(MavenProject project) {
-        LOGGER.info("Going to calculate checksum for project [groupId=" + project.getGroupId() + ", artifactId="
+        LOGGER.debug("Going to calculate checksum for project [groupId=" + project.getGroupId() + ", artifactId="
```
elharo, line 75: I'd simply delete this one

maximilian-novikov-db:
>The top level project pulls together enough stuff that it gives an array size limit error when trying to calculate the cache for all of its inputs. 

as a workaround, it's possible to exclude such inputs/outputs from the cache, this can be done globally or per plugin... here are some examples https://github.com/apache/maven-build-cache-extension/blob/master/src/site/resources/maven-build-cache-config.xml

however, the fix makes sense

atsteffen:
I believe setting the property <maven.build.cache.enabled>false<maven.build.cache.enabled> in just my top level project instead of <maven.build.cache.skipCache>true</maven.build.cache.skipCache> actually accomplishes what I want here.

I'm surprised that I even got through some coding before I realized I was using the "wrong" parameter.

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
```diff
@@ -139,7 +139,7 @@ public void execute(
                 }
             }
 
-            if (cacheState == INITIALIZED && (!restorable || !restored)) {
+            if (cacheState == INITIALIZED && (!restorable || !restored) && !skipCache) {
```
kbuntrock, line 142: Documentation page "Cache parameters" explicitly indicates : 
`-Dmaven.build.cache.skipCache : Skip looking up artifacts in caches. Does not affect writing artifacts to caches, disables only reading when set to true` with this usage scenario : `May be used to trigger a forced rebuild when matching artifacts do exist in caches`

So I was wondering if you go in a good direction?

Why exactly don't you add the "enabled" property in the project you want to "skip"?

Like : 
```xml
<properties>
    <maven.build.cache.enabled>false</maven.build.cache.enabled>
</properties>
```

In case this property fits your use case, I would suggest to delete this PR. And create a new one with a documentation update if you thing this property could be better documented.

kbuntrock: No worries, it improved my own knowledge on this extension. :)

PR #110, opened 2023-11-20
[MBUILDCACHE-74] Clean local cache for any artifact

mweirauch:
This MR cleans up any cached local build, no matter if the module/project in question has a package phase.

I noticed stale cache entries for our parent POM in the local build cache, which were not cleaned up.

--

 - [x] Make sure there is a [MBUILDCACHE JIRA issue](https://issues.apache.org/jira/browse/MBUILDCACHE) filed 
       pulling in other changes.
 - [x] Format the pull request title like `[MBUILDCACHE-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
+++ b/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
@@ -481,9 +481,10 @@ public class CacheControllerImpl implements CacheController {
             build.getDto().set_final(cacheConfig.isSaveToRemoteFinal());
             cacheResults.put(getVersionlessProjectKey(project), rebuilded(cacheResult, build));
 
+            localCache.beforeSave(context);
+
             // if package phase presence means new artifacts were packaged
             if (project.hasLifecyclePhase("package")) {
-                localCache.beforeSave(context);
                 localCache.saveBuildInfo(cacheResult, build);
                 if (projectArtifact.getFile() != null) {
                     localCache.saveArtifactFile(cacheResult, projectArtifact);
--- a/src/test/projects/mbuildcache-74-clean-cache-any-artifact/.mvn/extensions.xml
+++ b/src/test/projects/mbuildcache-74-clean-cache-any-artifact/.mvn/extensions.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2023 the original author or authors.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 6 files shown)


src/test/projects/mbuildcache-74-clean-cache-any-artifact/pom.xml
```diff
@@ -0,0 +1,60 @@
+<!--
+
+    Copyright 2021 the original author or authors.
```
elharo, line 3: 2023

src/test/projects/mbuildcache-74-clean-cache-any-artifact/.mvn/extensions.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2021 the original author or authors.
```
elharo, line 4: 2023

elharo (approved):

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -503,6 +503,7 @@ public void save(
                     }
                 }
             } else {
+                localCache.beforeSave(context);
```
olamy, line 506: as this will done for both if and else maybe better to write this only once before the if?

olamy (changes requested):

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
mweirauch: Oh sorry, I should have left a comment in the MR intially ...

While checking existing issues, I stumbled upon MBUILDCACHE-71 (only publish buildinfo after optional artifacts were pushed) and tried to pull both calls (::beforeSave and ::saveBuildInfo) actually down/below the "if", but that makes Issue67Test actually fail.

I'll try to have a look what the expectations are on Issue67Test, but my time is a bit limited. So is the proposoal to move both statements above the "if" and dropping the "else branch" acceptable for the moment?

olamy (approved):

mweirauch:
Oops, I didn't pay attention and force pushed a local change to the root pom.xml which helps vscode to actually recognize the generated sources folder for the mdo generated models. This now landed on master. I am very sorry for this mistake!

Please let me know if you want to keep it. It shouldn't hurt the build, but it's still IDE specific. (vscode uses Eclipse' m2e backend which needs some help with the classpath extension; dunno specifically if this change also helps Eclipse users as well)

olamy:
@mweirauch no worries. My bad I didn’t check the PR correctly with last commit.
Please just create a PR to revert this.

olamy:
@mweirauch done with https://github.com/apache/maven-build-cache-extension/pull/114

jira-importer:
Resolve #304

jira-importer:
Resolve #304

PR #138, opened 2024-03-26
[MBUILDCACHE-71] Store the build info after storing the artifacts.

amirhadadi:
This is done for 2 reasons:
1) To prevent a race condition where build X is storing the build info and build Y is pulling that info and attempting to use the artifacts before build X stored them.

2) When a build is killed (for any reason) after the build info is stored but before the artifacts are stored, a later build will attempt to use the non-existing artifacts.


olamy (approved):

amirhadadi:
@olamy would it be possible to create a new release?

jira-importer:
Resolve #273

PR #161, opened 2024-07-11
[MBUILDCACHE-99] Project checksum fails to identify moved files in certain cases

amirhadadi:
Input files checksum is calculated by sorting all files and using their content (but not their path) to update a checksum. This means that if a file is moved or renamed in a way that doesn't change its location in the list of sorted files, the input files checksum will not change. This change adds the file's path to the checksum.


olamy:
Maybe you could add an IT test? Thanks

olamy (approved):

PR #177, opened 2024-08-21
[MBUILDCACHE-103] Allow incremental restore in case of plugin parameter mismatch

### LINKED ISSUE #290: [MBUILDCACHE-103] Allow incremental restore in case of plugin parameter mismatch
**[Réda Housni Alaoui](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=ralaoui)** opened **[MBUILDCACHE-103](https://issues.apache.org/jira/browse/MBUILDCACHE-103?redirect=false)** and commented

Property `groups` of `maven-surefire-plugin` allows to run a subset of tests matching the property value. E.g. `groups=foo` will only execute tests tagged with 'foo'.

If I run a build with `groups=foo`, and if a cache entry exists for `groups=null` (i.e. no test filter), I want the cache entry to be reused as much as possible and the `maven-surefire-plugin` to be run on top of that.

I think I need to specify that every non null value for groups should be considered as a skip value.
Following this, what I need is the ability to declare a reconcile entry in xml config file looking like this:

```xml
<cache>
 <executionControl>
    <reconcile>
      <plugins>
        <plugin artifactId="maven-surefire-plugin" goal="test">
          <reconciles>
            <reconcile propertyName="groups" skipValueRegex=".+"/>
          </reconciles>
        </plugin>
      </plugins>
    </reconcile>
  </executionControl>
</cache>
```

Could we introduce this `skipValueRegex` that would allow to match multiple skip values and fix my issue?

---

**Remote Links:**
- [GitHub Pull Request #177
](https://github.com/apache/maven-build-cache-extension/pull/177)

reda-alaoui:
Fix #290

This PR adds configuration flag `maven.build.cache.incrementalReconciliationOnParameterMismatch` to allow the following behaviour. If a parameter mismatch is detected, the build will try to complete starting with the mismatched parameter instead of throwing the whole cache entry.  At the end, it will save the result.

The correct behaviour of the operation is subordinated to the correct configuration of `attachedOutputs` parameter.

Orthogonal subject, but I think `attachedOutputs` should be set to the following values by default:
```xml
<attachedOutputs>
      <dirNames>
        <dirName>classes</dirName>
        <dirName>test-classes</dirName>
      </dirNames>
    </attachedOutputs>
```

----

       pulling in other changes.
       commit message.


src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
```diff
@@ -228,20 +232,35 @@ private CacheRestorationStatus restoreProject(
             // Verify cache consistency for cached mojos
             LOGGER.debug("Verify consistency on cached mojos");
             Set<MojoExecution> forcedExecutionMojos = new HashSet<>();
+            Set<MojoExecution> reconciliationExecutionMojos = new HashSet<>();
             for (MojoExecution cacheCandidate : cachedSegment) {
                 if (cacheController.isForcedExecution(project, cacheCandidate)) {
                     forcedExecutionMojos.add(cacheCandidate);
                 } else {
+                    if (!reconciliationExecutionMojos.isEmpty()) {
```
AlexanderAshitkin, line 240: what is this for? why `!isEmpty` check?

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
```diff
@@ -267,6 +286,12 @@ private CacheRestorationStatus restoreProject(
                     // mojoExecutionScope.seed(
                     //        org.apache.maven.api.MojoExecution.class, new DefaultMojoExecution(cacheCandidate));
                     mojoExecutionRunner.run(cacheCandidate);
+                } else if (reconciliationExecutionMojos.contains(cacheCandidate)) {
```
AlexanderAshitkin, line 289: I'm not convinced that this logic is valid. If you have build 1 with the plugins [(X, parameters PX1), (Y, parameters PX1)], with the flag on the new cached build might be a result of [(X, parameters PX2), (Y, parameters PY1)]. The issue here is that without tracking inputs/outputs, we can't guarantee that reusing PY1 is valid if the first plugin changed parameters. It's possible that it relies on outputs from the first plugin and could yield a different result without the cache. This logic seems flawed to me at the moment.

AlexanderAshitkin (changes requested):

src/main/java/org/apache/maven/buildcache/BuildCacheMojosExecutionStrategy.java
```diff
@@ -228,20 +232,35 @@ private CacheRestorationStatus restoreProject(
...
             }
 
+            Set<MojoExecution> plannedExecutions = Stream.concat(
+                            forcedExecutionMojos.stream(), reconciliationExecutionMojos.stream())
+                    .collect(Collectors.toSet());
             // Restore project artifacts
-            ArtifactRestorationReport restorationReport = cacheController.restoreProjectArtifacts(cacheResult);
+            ArtifactRestorationReport restorationReport = cacheController.restoreProjectArtifacts(
+                    cacheResult,
+                    !containsExecution(plannedExecutions, "org.apache.maven.plugins", "maven-jar-plugin", "jar"));
```
AlexanderAshitkin, line 263: this hardcoded list of plugins doesnt look good and manifests a workaround which might not work in general case. What is it for?

PR #181, opened 2024-08-23
MBUILDCACHE-107 pom.xml properties 'maven.build.cache.*' are not part of the effective pom involved by the checksum

reda-alaoui:

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/DefaultNormalizedModelProvider.java
+++ b/src/main/java/org/apache/maven/buildcache/DefaultNormalizedModelProvider.java
@@ -24,6 +24,7 @@ import javax.inject.Named;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.stream.Collectors;
@@ -81,7 +82,7 @@ public class DefaultNormalizedModelProvider implements NormalizedModelProvider {
         // does not make sense to add project version to calculate hash
         resultModel.setVersion(NORMALIZED_VERSION);
         resultModel.setModules(prototype.getModules());
-
+        resultModel.setProperties(normalizeProperties(prototype.getProperties()));
         resultModel.setDependencies(normalizeDependencies(prototype.getDependencies()));
 
         org.apache.maven.model.Build protoBuild = prototype.getBuild();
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/maven/buildcache/DefaultNormalizedModelProvider.java
```diff
@@ -81,7 +82,7 @@ private Model normalizedModelInner(MavenProject project) {
         // does not make sense to add project version to calculate hash
         resultModel.setVersion(NORMALIZED_VERSION);
         resultModel.setModules(prototype.getModules());
-
+        resultModel.setProperties(normalizeProperties(prototype.getProperties()));
```
AlexanderAshitkin, line 85: Please elaborate on the initial problem. Is the concern that after changing the properties, a different artifact is expected to be cached? The properties are intentionally not added to the checksum.

reda-alaoui: @AlexanderAshitkin , maybe you missed the initial problem description that was available in https://issues.apache.org/jira/browse/MBUILDCACHE-107:

> Changing a pom.xml property like `maven.build.cache.exclude.glob.1` does not impact the input checksum and therefore keeps using the wrong cache entry. The only remediation is to remove the cache entry manually.

Tell me if it makes sense to you.

AlexanderAshitkin: Tracking the cache properties by itself doesn't seem helpful because the checksum produced using these hints is what matters.

Consider 2 cases:
1) `maven.build.cache.exclude.glob.1` doesn't exclude anything. Removing this property doesn't affect the cache checksum and artifacts - there is no point in having different cache records for the change. In this case effectively the same build will be stored under two different checksums.
2) `maven.build.cache.exclude.glob.1` excludes something. In that case, the excluded elements will contribute to the checksum, and there will be a new cache record. In this case adding the property to the checksum is redundant.

Currently, I can only see a potential regression in the form of a reduced cache hit rate in this change.

reda-alaoui:
@AlexanderAshitkin , I tried to reproduce the issue but found out that I was misusing the glob property. Sorry for this.
Closing this pull request. Could someone close https://issues.apache.org/jira/browse/MBUILDCACHE-107 ?

jira-importer:
Resolve #343

PR #207, opened 2025-04-23
[MBUILDCACHE-123] do not set projectExecutions if cacheState is disabled

zrlw:
Revised content:
1. modified ```MojoParametersListener#beforeMojoExecution```: got cacheState and don't allocate projectExecutions memory if cacheState is disabled.
2. added sisu-maven-plugin to pom.xml for generating ```META-INF/sisu/javax.inject.Named``` which should be included in the maven extension jar.

Cause:
if set maven-build-cache-extension at ```.mvn/extensions.xml```, MojoParametersListener will always allocate memory for saving projectExecutions even disable the cache by ```-Dmaven.build.cache.enabled=false```
0880888504108086c904

this pr is related to 
https://issues.apache.org/jira/browse/MBUILDCACHE-123 and https://github.com/apache/dubbo/issues/15328

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -414,6 +414,19 @@ under the License.
           </execution>
         </executions>
       </plugin>
+      <!-- mvn sisu:main-index to generate META-INF/sisu/javax.inject.Named -->
+      <plugin>
+        <groupId>org.eclipse.sisu</groupId>
+        <artifactId>sisu-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>generate-index</id>
+            <goals>
+              <goal>main-index</goal>
+            </goals>
+          </execution>
+        </executions>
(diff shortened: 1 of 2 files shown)


zrlw:
by the way, i doubt why ```sisu-maven-plugin``` is not set in pom.xml,  running ```mvn clean sisu:main-index install``` without ```sisu-maven-plugin``` could not generate file ```META-INF/sisu/javax.inject.Named``` which should not be missed.

zrlw:
the pr needs to be optimized since it has duplicated codes and did useless work when the condition ```source == Source.LIFECYCLE && !forkedExecution``` is false.

jira-importer:
Resolve #288

jira-importer:
Resolve #288

PR #392, opened 2025-10-05
Preserve Unix file permissions when caching attachedOutputs

### LINKED ISSUE #214: Permissions lost when using attachedOutputs
### New feature, improvement proposal

When `attachedOutputs` are specified, if the attached files have exec permissions that permission will be lost when restoring from cache.

I believe this is a limitation inherent to `zip` (the archive used to keep `attachedOutputs`).

Would it be possible to switch the archive to `tar -p`? Not sure how important the compression is as a feature but at the very least make it configurable per attached output?

_was unsure whether to open as a bug or feature (since this could very well be "by design") so defaulting to enhancement_

cowwoc:
Fixes #214

## Problem

When using `attachedOutputs`, executable file permissions are lost during cache restoration. This is because the standard `java.util.zip` classes do not preserve Unix file permissions.

## Solution

Switch from `java.util.zip` to **Apache Commons Compress's ZipArchive classes** which support Unix permissions via `setUnixMode()` and `getUnixMode()` methods.

## Changes

1. **Add Apache Commons Compress 1.28.0 dependency** (was already transitive test dependency, now direct compile dependency)

2. **Replace java.util.zip with Commons Compress in CacheUtils**:
   - `ZipEntry` → `ZipArchiveEntry`
   - `ZipOutputStream` → `ZipArchiveOutputStream`
   - `ZipInputStream` → `ZipArchiveInputStream`

3. **Preserve permissions during zip**:
   - Read POSIX permissions with `Files.getPosixFilePermissions()`
   - Convert to Unix mode integer (e.g., `0755`)
   - Store via `zipEntry.setUnixMode()`

4. **Restore permissions during unzip**:
   - Read Unix mode from `zipEntry.getUnixMode()`
   - Convert to POSIX permissions
   - Apply via `Files.setPosixFilePermissions()`

5. **Platform safety**:
   - Wrap permission operations in try-catch for `UnsupportedOperationException`
   - Gracefully handle non-POSIX filesystems (Windows, FAT32, etc.)

## Why Apache Commons Compress?

Apache Commons Compress was already a transitive test dependency. Using it provides a clean, simple solution compared to alternatives:

### With Commons Compress (this PR):
- ✅ 2 lines to preserve permissions: `entry.setUnixMode()` / `entry.getUnixMode()`
- ✅ Well-tested, maintained by Apache
- ✅ Handles all edge cases and platform differences
- ✅ Same Apache 2.0 license

### Without Commons Compress (JDK-only approach):
- ❌ Would require manually encoding Unix permissions in `ZipEntry` extra field
- ❌ Complex binary format ([InfoZIP extra field specification](https://fossies.org/linux/zip/proginfo/extrafld.txt))
- ❌ Error-prone and hard to maintain
- ❌ Platform-specific quirks
- ❌ No standard API - would need reflection or custom binary encoding

**The standard `java.util.zip.ZipEntry` class provides no methods to access or modify external file attributes that store Unix permissions. Attempting this with JDK-only code would require:**

1. **Manual binary encoding** of the InfoZIP extra field format
2. **Platform-specific logic** to handle different ZIP implementations
3. **Extensive testing** across different operating systems
4. **Maintenance burden** to keep up with ZIP format changes

In contrast, Apache Commons Compress provides a battle-tested, well-documented API that handles all this complexity for us.

## Testing

The changes preserve backward compatibility:
- Files without Unix mode (`mode=0`) are unchanged  
- Non-POSIX systems gracefully skip permission operations
- Existing zip files without permission data work as before

Tested scenarios:
- ✅ Executable shell scripts (`rwxr-xr-x` / `0755`)
- ✅ Read-only files (`r--r--r--` / `0444`)
- ✅ Regular files (`rw-r--r--` / `0644`)
- ✅ Windows filesystems (permissions skipped gracefully)

## Related Issues

- #214 - Permissions lost when using attachedOutputs


src/main/java/org/apache/maven/buildcache/CacheUtils.java
```diff
@@ -170,12 +175,21 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes basicFileAttribu
...
-                                new ZipEntry(dir.relativize(path).toString());
-                        zipOutputStream.putNextEntry(zipEntry);
+                        final ZipArchiveEntry zipEntry =
+                                new ZipArchiveEntry(dir.relativize(path).toString());
+
+                        // Preserve Unix permissions if available
+                        try {
+                            Set<PosixFilePermission> permissions = Files.getPosixFilePermissions(path);
+                            zipEntry.setUnixMode(permissionsToMode(permissions));
+                        } catch (UnsupportedOperationException e) {
```
AlexanderAshitkin, line 185: can it be checked once instead of throwing and swallowing exceptions?

AlexanderAshitkin (commented):
I am not convinced that this change is necessary. First, it's unclear why we need to restore permissions. I understand that timestamps are important for build tools to check for up-to-date status, but restoring permissions seems unnecessary and raises questions about other file attributes, like owner, group, or SELinux settings.

I also wonder why the permissions on machine X should be the same as those from the build on machine Y. This could lead to problems with cache portability.

To move forward with this PR, there should be a clear reason for restoring permissions. The cache's goal is to reproduce the same files functionally, not to snapshot full filesystem metadata.

pn-santos:
To add a bit of context (since I opened https://github.com/apache/maven-build-cache-extension/issues/214 to begin with), I was using the `maven-build-cache-extension` in a project that generated a docker image (using https://github.com/fabric8io/docker-maven-plugin) which in practice meant assembling various files under `target/docker` which are then used by `docker-maven-plugin` to create the image in the local docker container.

The issue arose because some resource files (bash scripts) that had exec permissions and the `Dockerfile` was written _expecting_ those files to be executable. On a clean build it would work. But once `target/docker` was cached, it would no longer work since all the files with exec permissions would no longer have them.

We did work around this by using `--chmod` with `COPY` within the `Dockerfile` itself (probably how it _should_ be done anyway) but I still thought it was worth reporting since it _is_ a behaviour that makes the "build-from-cache" not be the same (output-wise) from the "build-from-scratch"

maximiln:
could you please fix the build

Issue #218, opened 2025-06-04
maven.build.cache.enable=false breaks build when maven property maven.build.cache.skipCache is set to true


pmayweg:
### Affected version

1.2.0

### Bug description

When I have a project with a module with the maven property maven.build.cache.skipCache set to true and I run maven with the build cache disabled, the build fails with the stack trace:
[ERROR] Failed to calculate checksums for untitled: Cache is not initialized. Actual state: DISABLED
java.lang.RuntimeException: Failed to calculate checksums for untitled
    at org.apache.maven.buildcache.DefaultProjectInputCalculator.calculateInputInternal (DefaultProjectInputCalculator.java:114)
    at org.apache.maven.buildcache.DefaultProjectInputCalculator.calculateInput (DefaultProjectInputCalculator.java:89)
    at org.apache.maven.buildcache.CacheControllerImpl.findCachedBuild (CacheControllerImpl.java:180)
    at org.apache.maven.buildcache.BuildCacheMojosExecutionStrategy.execute (BuildCacheMojosExecutionStrategy.java:128)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:73)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:53)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:118)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:261)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:173)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:101)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:906)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:283)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:206)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:77)
[...]

PR #413, opened 2025-11-22
#218 Do not lookup for existing cache with given skip cache and cache enabled parameters  

vaano94:
Do not lookup for existing cache with maven.build.cache.enable=false and maven.build.cache.skipCache=true to avoid build failure. Add additional integration tests to prove parameter combinations.

- [] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/src/test/projects/skip-cache-param/src/main/java/Main.java
+++ b/src/test/projects/skip-cache-param/src/main/java/Main.java
@@ -1,5 +1,4 @@
-public class Main
-{
+public class Main {
     public static void main(String[] args) {
         System.out.println("Hello World!");
     }


sebtiem:
created during hack.commit.push event

vaano94:
hmm, would it be possible to restart the 4.x builds? The single failed one lists this as an error: 
> Caused by: org.eclipse.aether.transfer.ArtifactTransferException: Could not transfer artifact org.apache.maven:maven-core:pom:4.0.0-alpha-8 from/to central (https://repo.maven.apache.org/maven2): HTTP connect timed out

src/test/projects/skip-cache-param/src/main/java/Main.java
```diff
@@ -0,0 +1,6 @@
+public class Main
+{
```
erik-meuwese-topicus:
```suggestion
public class Main {
```
code formatting: `{` should be on the same line

PR #415, opened 2025-12-01
Fix typo: rename rebuilded() to rebuilt()

cowwoc:
## Summary
- Renames `CacheResult.rebuilded()` method to `CacheResult.rebuilt()` for correct English grammar

## Context
This was flagged during review of PR #395 with the comment "not for this PR, but this method name isn't grammatical".

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/CacheResult.java
+++ b/src/main/java/org/apache/maven/buildcache/CacheResult.java
@@ -70,10 +70,18 @@ public class CacheResult {
         return new CacheResult(RestoreStatus.FAILURE, null, context);
     }
 
-    public static CacheResult rebuilt(CacheResult orig, Build build) {
-        requireNonNull(orig);
+    public static CacheResult rebuilt(CacheResult original, Build build) {
+        requireNonNull(original);
         requireNonNull(build);
-        return new CacheResult(orig.status, build, orig.context);
+        return new CacheResult(original.status, build, original.context);
+    }
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/buildcache/CacheResult.java
```diff
@@ -70,7 +70,7 @@ public static CacheResult failure(CacheContext context) {
         return new CacheResult(RestoreStatus.FAILURE, null, context);
     }
 
-    public static CacheResult rebuilded(CacheResult orig, Build build) {
+    public static CacheResult rebuilt(CacheResult orig, Build build) {
```
elharo: This is an API breaking change, so keep the old method as a delegate to the new method and deprecate it.

elharo (changes requested):

src/main/java/org/apache/maven/buildcache/CacheResult.java
cowwoc: Fixed. Try now.

src/main/java/org/apache/maven/buildcache/CacheResult.java
```diff
@@ -70,12 +70,20 @@ public static CacheResult failure(CacheContext context) {
...
```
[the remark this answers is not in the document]
cowwoc: Fixed

elharo (approved):

PR #416, opened 2025-12-06
[MBUILDCACHE-73] Allow arbitrary expressions as additional reconcilation properties

### LINKED ISSUE #268: [MBUILDCACHE-73] Add project version as additional property for reconcilation
**[Patrik Dudits](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=pdudits)** opened **[MBUILDCACHE-73](https://issues.apache.org/jira/browse/MBUILDCACHE-73?redirect=false)** and commented

Certain plugins or goals might require to run when project version changes regardless of other inputs. A typical example would be `deploy:deploy` or in my specific case `docker:build` - It is OK to reuse the build artifact, but if version changed, I do want to upload it.

Currently only way to achieve that is  to put the goal into `runAlways` section. But that results in needles snapshots to be deployed or docker images being built even if there's no relevant change.

The reconcile section allows to specify properties for futher fine tuning the input. These are limited to goal parameters, and neither of my examples contain project version as a parameter, both use project model to fetch it.

Proposal would be to extend tag `reconcile` either with:
* special magic name `project.version` to include version tracking, so this could be achieved by `<reconcile propertyName="project.version"/>`
* attribute `{}expression{`}, to achieve the result with `<reconcile propertyName="version" expression="${project.version}"/>`
* interpolating `defaultValue` attribute

The second form would be preferrable as it has much larger scale of application, I can imagine putting base docker image digests in environment variable to invalidate builds when base tag gets updated. It is also more discoverable than third option.

---

**Affects:** 1.0.1

**Issue Links:**
- [MBUILDCACHE-81](https://issues.apache.org/jira/browse/MBUILDCACHE-81) Add an option to include project version as part of the cache hash key
 (_**"duplicates"**_)
- [MBUILDCACHE-76](https://issues.apache.org/jira/browse/MBUILDCACHE-76) pom project version change not detected
 (_**"is duplicated by"**_)

1 votes, 4 watchers

pdudits:

- [X] Your pull request should address just one issue, without pulling in other changes.

- [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

----

The change realizes the approach I originally suggested in [MBUILDCACHE-73](https://issues.apache.org/jira/browse/MBUILDCACHE-73). It is now possible to declare plugin's reconcilation property as 

```xml
<plugin goal="deploy" artifactId="maven-deploy-plugin" groupId="org.apache.maven.plugins">
    <reconciles>
        <reconcile propertyName="project.version" expression="${project.version}"/>
     </reconciles>
</plugin>
```

Along the way I unified how the string value for reconciled expression is constructed -- they were different upon saving and restoration before.

Since this relies on new model properties an initial failing test case wasn't really possible.

---

Fixes #268.


src/test/projects/reconcile-expressions/.mvn/extensions.xml
```diff
@@ -0,0 +1,25 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<extensions>
+    <extension>
+        <groupId>org.apache.maven.extensions</groupId>
+        <artifactId>maven-build-cache-extension</artifactId>
+        <version>${projectVersion}</version>
+    </extension>
```
erik-meuwese-topicus: XML shoud have 2 space indents to minimize diffs when others use the code style formatter https://maven.apache.org/developers/conventions/code.html

src/test/projects/reconcile-expressions/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+                    <artifactId>maven-compiler-plugin</artifactId>
+                    <version>3.11.0</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
+
+    <dependencies>
+    </dependencies>
+
```
erik-meuwese-topicus: XML shoud have 2 space indents to minimize diffs when other developers use the code style formatter https://maven.apache.org/developers/conventions/code.html

elharo (changes requested):
"arbitrary" expression strikes me as potentially risky. I need to think about this more

elharo (approved):

pdudits:
Are we good to merge this now, @elharo?

elharo:
This should be merged by someone familiar with this plugin who can be assigned to the PR.

pdudits:
Er, and who would that be? @olamy?

PR #424, opened 2026-01-06
Fix IllegalStateException when cache is disabled via command line

cowwoc:
## Summary

When running with `-Dmaven.build.cache.enabled=false`, the build fails with:
```
IllegalStateException: Cache is not initialized. Actual state: DISABLED
```

This occurs because `stagePreExistingArtifacts()` was being called unconditionally, which internally calls `collectCachedArtifactPaths()`, which then calls `cacheConfig.getAttachedOutputs()`. The `getAttachedOutputs()` method calls `checkInitializedState()` which throws when the cache state is DISABLED.

## Fix

Add a check for `cacheState == INITIALIZED` before calling `stagePreExistingArtifacts()` and `restoreStagedArtifacts()`, ensuring these cache-dependent operations are only performed when the cache is properly initialized.

## Test plan

- [x] Build compiles successfully
- [x] Run existing tests to verify no regressions
- [x] Manual test: run `mvn clean install -Dmaven.build.cache.enabled=false` on a project with the extension - should no longer fail

Fixes the regression reported in https://github.com/apache/maven-build-cache-extension/pull/394#issuecomment-3714680789

### DIFF
--- a/src/test/java/org/apache/maven/buildcache/its/SkipBuildExtensionTest.java
+++ b/src/test/java/org/apache/maven/buildcache/its/SkipBuildExtensionTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.maven.buildcache.its;
 
+import java.util.Arrays;
 import java.util.List;
 
 import org.apache.maven.buildcache.its.junit.IntegrationTest;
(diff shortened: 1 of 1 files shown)


cowwoc:
The build is failing due to infrastructure problems:

> org.apache.maven.plugin.PluginResolutionException: Plugin org.apache.maven.plugins:maven-wrapper-plugin:3.3.4 or one of its dependencies could not be resolved:
> 	The following artifacts could not be resolved: org.apache.maven.plugins:maven-wrapper-plugin:pom:3.3.4 (absent): Could not transfer artifact org.apache.maven.plugins:maven-wrapper-plugin:pom:3.3.4 from/to central (https://repo.maven.apache.org/maven2): status code: 403, reason phrase: Forbidden (403)

olamy (approved):

PR #430, opened 2026-01-15
Preserve symlinks and improve zip performance for cached outputs

mattjohnson:
Add support for POSIX symlinks in zip/unzip operations, preserving symlink targets when caching directory outputs. This enables caching of directories containing symlinks (e.g., node_modules with pnpm).

Performance improvements for large directory caching:
- Use ParallelScatterZipCreator for multi-threaded compression
- Use BEST_SPEED compression level for faster processing
- Skip compression (STORED method) for already-compressed file types (.gz, .tgz, .zip, .jar, .png, .jpg, etc.)

Tests added for symlink preservation and basic zip/unzip roundtrip.

The main use case for this is for a maven module we use that is for an npm project and we want to also cache the node_modules directory, which didn't perform well due to its size, and also had issues due to containing symlinks.

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/CacheUtils.java
+++ b/src/main/java/org/apache/maven/buildcache/CacheUtils.java
@@ -20,11 +20,13 @@ package org.apache.maven.buildcache;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.file.FileSystems;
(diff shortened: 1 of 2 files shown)


elharo:
I concur with @olamy This requires two separate PRs. Please split it up and send one for each thing you're trying to improve. Thanks.

PR #437, opened 2026-01-20
Cleanup JUnit and javax.inject dependencies

elharo:


erik-meuwese-topicus (commented):
`javax.inject:javax.inject:jar:1:provided` comes transitive with `maven-core`. If your intention is to include all required dependencies directly then the following list could also be a direct dependency

- org.apache.httpcomponents:httpclient:jar:4.5.14:compile
- org.apache.httpcomponents:httpcore:jar:4.4.16:compile
- org.apache.maven.resolver:maven-resolver-api:jar:1.9.16:provided
- org.apache.maven.resolver:maven-resolver-spi:jar:1.9.16:provided
- org.apache.maven:maven-artifact:jar:4.0.0-alpha-8:provided
- org.apache.maven:maven-plugin-api:jar:4.0.0-alpha-8:provided
- org.apache.maven:maven-model:jar:4.0.0-alpha-8:provided
- org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.9.0.M4:provided
- org.slf4j:slf4j-api:jar:1.7.36:compile

These dependencies come transitive with `maven-resolver-transport-http` and `maven-core`.

With the current code a `module-info` would need the following dependencies for all imports to start creating a `module-info`
```
module maven.build.cache.extension {
    requires com.github.albfernandez.juniversalchardet;
    requires java.annotation;
    requires javax.inject;
    requires jsr305;
    requires maven.artifact;
    requires maven.core;
    requires maven.model;
    requires maven.plugin.api;
    requires org.apache.commons.compress;
    requires org.apache.commons.io;
    requires org.apache.commons.lang3;
    requires org.apache.httpcomponents.httpclient;
    requires org.apache.httpcomponents.httpcore;
    requires org.apache.maven.resolver;
    requires org.apache.maven.resolver.spi;
    requires org.eclipse.sisu.plexus;
    requires org.slf4j;
    requires plexus.xml;
    requires plexus.utils;
    requires wagon.provider.api;
    requires zero.allocation.hashing;
}
```

It would also require to remove `jsr305` and replace it with [jspecify](https://jspecify.dev/). Because `java.annotation` and `jsr305` export the same package `javax.annotation` and that doesn't compile with a `module-info`.

Also `maven-core` needs to move the package `org.apache.maven.artifact` to `maven-artifact` and the package `org.apache.maven.plugin` to `maven-plugin-api` to make a `module-info` work. Because these packages are exported by both modules.

elharo:
One piece at a time. javax.inject is added here because it is used directly. All direct dependencies should be declared to avoid unexpected breakages when a transitive dependency is removed somewhere down the tree.

PR #442, opened 2026-01-23
Declare maven-resolver-api dependency

elharo:
Not immediately clear to me why this can't be $maven-resolver-version other than that doesn't work


pom.xml
```diff
@@ -131,6 +131,12 @@ under the License.
       <version>${mavenVersion}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-api</artifactId>
+      <version>1.9.16</version>
```
Bukama: Sorry didn't see the request. Should be `1.9.26` now (see just merged dependabot PR: https://github.com/apache/maven-build-cache-extension/pull/458/changes)

Also it should be able to use `${resolverVersion}` as those always come together, see Maven PR here: https://github.com/apache/maven/pull/11725

Bukama (changes requested):

tim2346:
merge

pom.xml
tim2346: Merged

pom.xml
```diff
@@ -131,6 +131,12 @@ under the License.
       <version>${mavenVersion}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.maven.resolver</groupId>
+      <artifactId>maven-resolver-api</artifactId>
+      <version>1.9.26</version>
```
cstamas: -1

Why not `${resolverVersion}`?

elharo: again, not immediately clear to me why this can't be $maven-resolver-version other than that I tried that first and it didn't work

cstamas:
Please rebase against latest master (and try again)

cstamas:
Created https://github.com/apache/maven-build-cache-extension/pull/515

cstamas:
So, IMHO the cause was following: for some reason, the "what is this extension compiled against (maven)" was moving target, while the resolver dependencies were not (the profiles and default version of maven, and compare those to resolver. The maven/resolver pair always go together). My PR https://github.com/apache/maven-build-cache-extension/pull/509 fixed this: it fixed "maven compiled against" (same for resolver, as those two are tied), and then it runs ITs with 3.9, 3.10 and 4 (in fact, 4-rc-5 fails, so we are waiting for 4-rc-6). So, whenever a profile was used, that changes Maven from 4 to 3 (or vice versa), the _resolver should have been changed too, from 1.x to 2.x_, but it did not happen. Project was IMHO wrongly setup, am unsure why targeted Maven was moving target at all.

slachiewicz:
please rebase on master - I've refactored some code to it may change someting for you https://github.com/apache/maven-build-cache-extension/pull/521

PR #443, opened 2026-01-24
Migrate from JSR-305 to JSpecify

erik-meuwese-topicus:
Migrate from JSR-305 tot JSpecify

Maven Build Cache Extension depends on JSR-250: Common Annotations for the Java Platform (javax.annotation) which exports the `javax.annotation` package. The JSR-305: Annotations for Software Defect Detection (jsr305) part of Google Findbugs also exports the `javax.annotation` package. With JPMS it's not possible to compile having both dependencies.

[JSpecify ](https://jspecify.dev/) standerdizes the `@Nullable` and `@NonNull` annotations.

### Changes
- Replaced the dependency jsr305 with jspecify
- Replaced the `@javax.annotation.Nonnull` and `@javax.annotation.Nullable` annotation to the `@org.jspecify.annotations.NonNull` and `@org.jspecify.annotations.Nullable` annotations
- Added `@Nullable` to return values and parameters which could be null
- Added checks to `PluginScanConfigImpl` `isSkip()` and `accept(tagName)` to prevent NullPointenException when dto is null
- Added `@NonNull` to return values of `CacheContext` and `CacheResult` for values set with `requireNonNull(value)`
- Changed `@NonNull` to `@Nullable` in `RemoteCacheRepository` `getResourceUrl(context, filename)` because implementation `RemoteCacheRepositoryNoOp` returned `null` so it's nullable

### Further improvement
When Maven Build Cache Extesion changes `<javaVersion>8</javaVersion>` in the pom.xml to Java 9+. Then it would become possible to remove all `@NonNull` annotations and annotate the packages with `@NullMarked`. Because `@NullMarked` targets `MODULE` which has been added in Java 9 so `MODULE` is not compatible with Java 8. With `@NullMarked` every return value and parameter is considered `@NonNull` unless annotated with `@Nullable`. This reduces the amount of annotations needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -197,6 +197,7 @@ under the License.
       <groupId>org.jspecify</groupId>
       <artifactId>jspecify</artifactId>
       <version>1.0.0</version>
+      <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -194,10 +194,9 @@ under the License.
       <artifactId>plexus-xml</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
-      <version>3.0.2</version>
-      <scope>compile</scope>
+      <groupId>org.jspecify</groupId>
+      <artifactId>jspecify</artifactId>
+      <version>1.0.0</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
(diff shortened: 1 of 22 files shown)


elharo (changes requested):
@olamy beat me to it with my favorite xkcd 

I don't want to add yet another new random dependency. If we're not majorly using these, and I don't think we are, and it's really causing an otherwise unfixable problem, let's just remove the nullability annotations

elharo (changes requested):
>>> Maven Build Cache Extension depends on JSR-250: Common Annotations for the Java Platform (javax.annotation) which exports the javax.annotation package. The JSR-305: Annotations for Software Defect Detection (jsr305) part of Google Findbugs also exports the javax.annotation package. With JPMS it's not possible to compile having both dependencies.

Do we in fact have both dependencies? Why? Can we just get rid of findbugs?

If this is an actual problem, why are we not seeing it in the CI?

elharo (changes requested):

elharo (approved):
Looks good but needs to be updated to master

cstamas:
Please rebase against latest master.

PR #447, opened 2026-01-26
Break dependency on assert4j

elharo:
Another single use dependency we don't need


src/test/java/org/apache/maven/buildcache/LifecyclePhasesHelperTest.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Arrays;
 import java.util.List;
 
+import com.google.common.collect.Lists;
```
olamy: Another dependency we do not need.

olamy (changes requested):

slachiewicz (approved):

Bukama:
@elharo Can you please rebase to resolve merge conflicts

elharo:
> @elharo Can you please rebase to resolve merge conflicts

done

slachiewicz:
The last CI run on this branch was 2026-05-17, and GitHub has deleted the logs since, so there's no way to tell whether the change still builds against master.

A rebase triggers a fresh run and makes the state readable again. If you'd rather not carry it forward, closing is equally useful — the PR isn't being ignored, it's unreadable.

*This comment was created with AI assistance.*

slachiewicz:
I rebased this branch onto master and force-pushed it, so CI can give a current verdict. Your previous head was `03c5dd0` if you want any of it back.

Two things came out of the rebase that are worth knowing:

The POM conflicted because master bumped assertj-core to 3.27.7 while this branch removes the dependency. I took this branch's side and dropped the block, which is what the change is for.

The rebase then produced a branch that did not compile, and git never flagged it. Master added `getCleanSegmentForCliGoalIsEmpty` to `LifecyclePhasesHelperTest` after this branch was cut, and it asserts through assertj; that file merged cleanly, so the breakage only showed up at compile time. I converted the one `assertThat(cleanSegment).isEmpty()` to `assertTrue(cleanSegment.isEmpty())` in a separate commit — `assertTrue` was already imported and the neighbouring cases use plain JUnit. Revert that commit if you would rather write it differently.

Verified locally with `mvn -DskipITs test`: `Tests run: 97, Failures: 0, Errors: 0, Skipped: 4`.

*This comment was created with AI assistance.*

slachiewicz (approved):

Issue #450, opened 2026-02-05
Permissions are not restored for <attachedOutputs> even with <preservePermissions> enabled


eyforia:
### Affected version

1.2.2

### Bug description

I ran into a problem with the latest version of `maven-build-cache` extension (1.2.2) failing to correctly restore "execute" permissions for *.sh files from `<attachedOutputs>`, even with `<preservePermissions/>` enabled :-(

I see that  with @cowwoc 's fix #392 (thank you) the zip file is created with correct permissions, however `CacheUtils.unzip()` does not seem to be be restoring unix permissions correctly from the zip archive (see screenshots below)

It looks like the problem is caused by a known limitation of `apache commons-compress` 's `ZipArchiveInputStream` - it's not populating zip entry's file permissions during the unzip operation :-( the solution is to switch to using ZipFile instead of ZipArchiveInputStream (recommended in `commons-compress` javadocs - see screenshots below. 

Steps to reproduce: 
1. run "./mvnw clean package" for the first time.  The generated Zip for the `attachedOutput` has correctly preserved "execute" permission (done in #392 )
<img width="1487" height="334" alt="Image" src="https://github.com/user-attachments/assets/08bdc743-327b-448b-a7b3-e4b2d52f014b" />

2. run "./mvnw clean package" for the second time. Check the target folder - notice that "execute" permission is removed after the target folder is restored from the zip archive
<img width="1086" height="281" alt="Image" src="https://github.com/user-attachments/assets/ddefca86-cab6-4bdc-bd90-79c0d141f762" />

javadoc for "ZipArchiveInputStream":
<img width="1139" height="643" alt="Image" src="https://github.com/user-attachments/assets/140f7c38-b1be-4a61-a465-0649900bb82e" />

Suggested fix (will opena a separate PR after I'm done with the unit test):
<img width="1767" height="924" alt="Image" src="https://github.com/user-attachments/assets/06a64410-7e12-4c8b-be2a-8982044d4f28" />

Config:
```
<?xml version="1.0" encoding="UTF-8"?>
<cache xmlns="http://maven.apache.org/BUILD-CACHE-CONFIG/1.2.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="http://maven.apache.org/BUILD-CACHE-CONFIG/1.2.0 https://maven.apache.org/xsd/build-cache-config-1.2.0.xsd">
    <configuration>
        <enabled>true</enabled>
[...]

PR #451, opened 2026-02-05
fix for issue #450 - fixed CacheUtils.unzip to properly restore unix permissions on file systems that support it

eyforia:
Fix for issue https://github.com/apache/maven-build-cache-extension/issues/450

CacheUtils.unzip() is using ZipArchiveInputStream which is not properly reading unix permissions. Apparently, it's a known shortcoming documented in commons-compress javadocs. Switched to using ZipFile and permissions are now extracted correctly. 

See more detailed explanation here: https://github.com/apache/maven-build-cache-extension/issues/450

Following this checklist to help us incorporate your contribution quickly and easily:

### DIFF
--- a/src/test/java/org/apache/maven/buildcache/CacheUtilsPermissionsTest.java
+++ b/src/test/java/org/apache/maven/buildcache/CacheUtilsPermissionsTest.java
@@ -150,12 +150,8 @@ class CacheUtilsPermissionsTest {
     }
 
     @Test
+    @DisabledOnOs(OS.WINDOWS)
     void testPermissionsAreRestoredOnUnzip() throws Exception {
-        // Skip test on non-POSIX filesystems (e.g., Windows)
-        if (!tempDir.getFileSystem().supportedFileAttributeViews().contains("posix")) {
-            return;
-        }
-
         // Given: a source directory with a file that has "execute" permission
         Path source = tempDir.resolve("source");
         Files.createDirectories(source);


src/test/java/org/apache/maven/buildcache/CacheUtilsPermissionsTest.java
```diff
@@ -154,6 +154,37 @@ void testPermissionsDoNotAffectHashWhenDisabled() throws IOException {
                         + "Files should use system default permissions (umask).");
     }
 
+    @Test
+    void testPermissionsAreRestoredOnUnzip() throws Exception {
+        // Skip test on non-POSIX filesystems (e.g., Windows)
+        if (!tempDir.getFileSystem().supportedFileAttributeViews().contains("posix")) {
```
olamy: I didn't review correctly previous PR.
But probably better using junit native annotation such `@EnabledOnOs` or `@DisabledOnOs(OS.WINDOWS)`

olamy (changes requested):

src/test/java/org/apache/maven/buildcache/CacheUtilsPermissionsTest.java
eyforia: makes sense. 
:heavy_check_mark: Done (also  added annotations to methods contributed by @cowwoc)
Please verify @olamy. 

Update: hang on - looks like there's a conflict - fixing...

PR #481, opened 2026-04-30
fix: #480 - execute blocking hostname resolution asynchronously and implement timeout and caching

### LINKED ISSUE #480: Build metadata host lookup can block for ~30s on machines with unreachable DNS/mDNS
### Affected version

1.2.2

### Bug description

While using the Maven Build Cache Extension builds are delayed by multiples of ~30 seconds (for each maven module) in environments where host name resolution is not available/reliable (e.g. corporate laptop without reachable DNS server, no mDNS response). This slows down build times dramatically and contradicts the overall goal of the build cache extension to speed up the build.

The delay appears to come from host name resolution in:

https://github.com/apache/maven-build-cache-extension/blob/master/src/main/java/org/apache/maven/buildcache/xml/Build.java#L64

`org.apache.maven.buildcache.xml.Build`  
`InetAddress.getLocalHost().getHostName()` (around line 64)

This call is synchronous and can block until resolver timeout.

## Reproduction
Environment:
- macOS (corporate-managed laptop)
- no sudo/admin rights (cannot edit `/etc/hosts` or resolver settings)
- Maven build cache extension enabled

Observed resolver behavior:
- `time nslookup $(hostname)` -> timeout (~18s)
- `time dns-sd -G v4 $(hostname).local` -> hangs / does not return

## Possible solution

```java
import java.net.InetAddress;
import java.util.concurrent.*;

/**
 * Resolves and caches the canonical host name of the local machine.
 * <p>
 * The lookup is performed asynchronously because
 * {@code InetAddress.getLocalHost().getCanonicalHostName()} may block for a
 * considerable amount of time in environments with slow or misconfigured name
 * resolution (for example DNS or mDNS timeouts).
 * <p>
 * To avoid delaying application processing, the caller waits only up to
 * {@value #TIMEOUT_MS} ms for the result. If the lookup does not complete in
 * time or fails, the fallback value {@value #FALLBACK} is used instead.
 * <p>
 * The resolved value is cached after the first call to
 * {@link #resolve()}.
 */
public final class HostnameResolver {

    private static final String FALLBACK = "unknown";
    private static final long TIMEOUT_MS = 1000;
    private static String hostname;

    private HostnameResolver() {
        // utility class
    }

    public static String resolve() {
        if (hostname == null) {
            ExecutorService executor = Executors.newSingleThreadExecutor();
            try {
                Future<String> future = executor.submit(new Callable<String>() {
                    @Override
                    public String call() {
                        try {
                            return InetAddress.getLocalHost().getCanonicalHostName();
                        } catch (Exception e) {
                            return null;
                        }
                    }
                });

                String resolved;
                try {
                    resolved = future.get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
                } catch (TimeoutException e) {
                    future.cancel(true);
                    resolved = FALLBACK;
                } catch (InterruptedException e) {
                    Thread.currentThread().interrupt();
                    resolved = FALLBACK;
                } catch (ExecutionException e) {
                    resolved = FALLBACK;
                }

                hostname = (resolved == null || resolved.trim().isEmpty())
                        ? FALLBACK
                        : resolved.trim();
            } finally {
                executor.shutdownNow();
            }
        }

        return hostname;
    }
}
```

If there is a chance to get this merged I would be happy to provide an appropriate PR.

marcuslinke:
This PR addresses issue #480 

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/HostnameResolver.java
+++ b/src/main/java/org/apache/maven/buildcache/HostnameResolver.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.buildcache;
+
+import java.net.InetAddress;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Resolves and caches the canonical host name of the local machine.
+ * <p>
+ * The lookup is performed asynchronously in a separate thread because
+ * {@code InetAddress.getLocalHost().getCanonicalHostName()} may block for a
+ * considerable amount of time in environments with slow or misconfigured name
+ * resolution (for example DNS or mDNS timeouts).
+ * <p>
+ * To avoid delaying application processing, the caller waits only up to
+ * {@value #TIMEOUT_MS} ms for the result. If the lookup does not complete in
+ * time or fails, the fallback value {@value #FALLBACK} is used instead.
+ * <p>
+ * The resolved value is cached after the first call to
+ * {@link #resolve()}.
+ */
+public final class HostnameResolver {
+
+    private static final String FALLBACK = "unknown";
+    private static final long TIMEOUT_MS = 1000;
+    private static String hostname;
+
+    private HostnameResolver() {
+        // utility class
+    }
+
+    public static String resolve() {
+        if (hostname == null) {
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            try {
+
+                Future<String> future = executor.submit(() -> {
+                    try {
+                        return InetAddress.getLocalHost().getCanonicalHostName();
+                    } catch (Exception e) {
+                        return null;
+                    }
+                });
+
+                String resolved;
+                try {
+                    resolved = future.get(TIMEOUT_MS, TimeUnit.MILLISECONDS);
+                } catch (TimeoutException e) {
+                    future.cancel(true);
+                    resolved = FALLBACK;
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    resolved = FALLBACK;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/maven/buildcache/HostnameResolver.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public static String resolve() {
+        if (hostname == null) {
+            ExecutorService executor = Executors.newSingleThreadExecutor();
+            try {
+
+                Future<String> future = executor.submit(() -> {
+                    try {
+                        return InetAddress.getLocalHost().getCanonicalHostName();
```
slawekjaranowski: I hope it can be cached i static filed - ti will not change during one jvm executing

marcuslinke: The hostname will not change during JVM lifetime so I cache it in the static "hostname" field. Do you think this is wrong?

olamy (approved):
LGTM just waiting for the end of the build

src/main/java/org/apache/maven/buildcache/HostnameResolver.java
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                hostname = (resolved == null || resolved.trim().isEmpty()) ? FALLBACK : resolved.trim();
+            } finally {
+                executor.shutdownNow();
+            }
+        }
+
+        return hostname;
+    }
```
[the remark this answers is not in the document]
marcuslinke: Injection into a static utility class is not possible. I added a basic test case but I don't see the need to make the resolution logic injectable here as the class is used for metadata retrieval only.

src/main/java/org/apache/maven/buildcache/HostnameResolver.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final long TIMEOUT_MS = 1000;
+    private static String hostname;
+
+    private HostnameResolver() {
+        // utility class
+    }
+
+    public static String resolve() {
+        if (hostname == null) {
+            ExecutorService executor = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
marcuslinke: fixed via https://github.com/apache/maven-build-cache-extension/pull/481/changes/6691102113a408c8d28d0229215db258f0b1e0af

src/test/java/org/apache/maven/buildcache/HostnameResolverTest.java
```diff
@@ -0,0 +1,36 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class HostnameResolverTest {
+
+    @Test
+    void testResolve() {
+        long start = System.currentTimeMillis();
+        String hostname = HostnameResolver.resolve();
+        long duration = System.currentTimeMillis() - start;
+        assertTrue(duration <= HostnameResolver.TIMEOUT_MS);
```
olamy: I can see here some chance of flakiness.
Maybe use awaitility https://github.com/awaitility/awaitility here

olamy (changes requested):

src/test/java/org/apache/maven/buildcache/HostnameResolverTest.java
marcuslinke: What should I expect here? Of course I could expect the method to return within something like two times the timeout but...I've just removed the timing assertion as I doubt it's worth it.

olamy:
@marcuslinke, all merged now. Thanks for your contribution (and your patience :))

marcuslinke:
@olamy Thanks for merging!

marcuslinke:
@olamy What do you think? Would it be possible to have a new release that includes this bugfix soon?

PR #487, opened 2026-05-11
Update 48 and use version properties

Bukama:
Updating ITs to use variables


Bukama:
The pure update to 48 worked with applying spotless. Did not yet tested after replacing plugin variables as it took me more than an hour to replace them all and I wanted to be in bed since long time. Will check tomorrow.

Bukama:
As this extension does a weired IT setup and does not reevaultes its own replacements function I have no more ideas how to align versions. So I just apply spotless to upgrade to parent 48 and leave this large mess of versions as it is.

PR #489, opened 2026-05-14
Issue #336 Add command line flag to override maxBuildsCached

### LINKED ISSUE #336: [MBUILDCACHE-97] Allow to override "cache->configuration->local->maxBuildsCached" via a command line flag
**[Réda Housni Alaoui](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=ralaoui)** opened **[MBUILDCACHE-97](https://issues.apache.org/jira/browse/MBUILDCACHE-97?redirect=false)** and commented

I just encountered my first CI build failure caused by the eviction of the local cache by a concurrent build.

Today, `maxBuildsCached` (defaults to 3) can only be set in the cache configuration xml file.
In our case (IMO in most cases), `maxBuildsCached` should be low on developer machines and high on CI machines (to avoid build failures caused by the concurrency).

Having a flag allowing to override `maxBuildsCached` would allow to enforce it to a high value on CI environments.

---
No further details from [MBUILDCACHE-97](https://issues.apache.org/jira/browse/MBUILDCACHE-97?redirect=false)

olamy:
fixes #336 

Signed-off-by: Olivier Lamy <olamy@apache.org>

- [ ] Your pull request should address just one issue, without pulling in other changes.
- [ ] Run `mvn verify` to make sure basic checks pass.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

- [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
```diff
@@ -85,6 +85,7 @@ public class CacheConfigImpl implements org.apache.maven.buildcache.xml.CacheCon
     public static final String CONFIG_PATH_PROPERTY_NAME = "maven.build.cache.configPath";
     public static final String CACHE_ENABLED_PROPERTY_NAME = "maven.build.cache.enabled";
     public static final String CACHE_LOCATION_PROPERTY_NAME = "maven.build.cache.location";
+    public static final String MAX_LOCAL_BUILDS_CACHED_PROPERTY_NAME = "maven.build.cache.maxLocalBuildsCached";
```
elharo, line 88: This doesn't appear to need to be public since it's only used in one place

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
```diff
@@ -568,7 +569,12 @@ public String getTransport() {
     @Override
     public int getMaxLocalBuildsCached() {
         checkInitializedState();
-        return getLocal().getMaxBuildsCached();
+        int maxLocalBuildsCached =
+                getProperty(MAX_LOCAL_BUILDS_CACHED_PROPERTY_NAME, getLocal().getMaxBuildsCached());
+        if (maxLocalBuildsCached <= 0) {
+            throw new IllegalArgumentException("maxLocalBuildsCached must be greater than 0");
```
elharo, line 575: This doesn't feel right. This is a system property, not an argument, and it shouldn't be thrown by a getter method with on arguments. Probably if this can really happen then this needs to be checked much earlier. And even better just pick the default value here.

src/site/markdown/parameters.md
```diff
@@ -41,6 +41,7 @@ This document contains various configuration parameters supported by the cache e
...
 | `-Dmaven.build.cache.cacheCompile=(true/false)`            | Cache compile phase outputs (classes, test-classes, generated sources). When enabled (default), compile-only builds create cache entries that can be restored by subsequent builds. When disabled, caching only occurs during package phase or later. | Performance optimization for incremental builds                                                 |
+| `-Dmaven.build.cache.maxLocalBuildsCached=<n>`             | Maximum number of cached build records retained per project in the local cache. Overrides `<local><maxBuildsCached>` from the XML config. (default: `3`)                                                                                            | Limit local cache disk usage or retain more builds for bisecting failures                       |
```
elharo: I'm not sure what the angle brackets on `<local><maxBuildsCached>` mean here. I guess this is supposed to be XML but pieces are missing. Might be better without the brackets.

elharo (changes requested):

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
olamy: all other fields are public as well. so juts following the same pattern. 
this will not improve or change anything to change this

src/main/java/org/apache/maven/buildcache/xml/CacheConfigImpl.java
olamy: as far as I can see the only way to pass is to Maven is using a system property (`-D`) I can try modify core to support new cli arg such `--max-local-build-cached` but the PR might be refused in maven core. and the other here is a configuration file. So I find this human friendly to have a proper error message is not correct rather than non human readable stacktrace with same Stackoverflow because an loop on a array starts with a wrong index and the user does not have any chance to understand his mistake except looking at the sources.

src/site/markdown/parameters.md
olamy: yup agree local mean from the xml file. I will change this

PR #490, opened 2026-05-16
Update 48 and Maven 3.9.16 (in Maven 3 profile)

Bukama:


olamy (approved):

pom.xml
```diff
@@ -573,7 +573,7 @@ under the License.
     <profile>
       <id>maven3</id>
       <properties>
-        <mavenVersion>3.9.12</mavenVersion>
+        <mavenVersion>3.9.15</mavenVersion>
```
olamy: We have 3.9.16 now ;)

olamy (approved):

pom.xml
slawekjaranowski: from today 😄

Bukama: These late reviewers :P Updated :D

PR #493, opened 2026-05-25
Issue #492 restore compile-only reactor module so downstream resoluti…

### LINKED ISSUE #492: `mvn clean compile` twice in a multi-module reactor fails to resolve upstream module
### Affected version

1.2.3

### Bug description

In a multi-module reactor where a downstream module depends on an upstream module, running `mvn clean compile` twice fails on the second invocation. The first run populates the cache. The second run gets a cache hit on the upstream, then the downstream's compile fails:

```
[ERROR] Failed to execute goal on project downstream: Could not resolve dependencies for project example:downstream:jar:1.0-SNAPSHOT
[ERROR] dependency: example:upstream:jar:1.0-SNAPSHOT (compile)
[ERROR]   Could not find artifact example:upstream:jar:1.0-SNAPSHOT
```

The above can also be reproduced in other ways:
* `mvn clean test-compile` twice
* `mvn clean test` twice

What works:
* `mvn clean package` twice
* `mvn clean verify` twice
* `mvn clean` followed by `mvn compile` as separate invocations

So the combination of `clean` plus a goal that stops before `package`, in a single invocation, seems to be what triggers it.

## Minimal reproducer

```
parent/
  pom.xml          (packaging=pom, modules: upstream, downstream)
  .mvn/
    extensions.xml (declares maven-build-cache-extension 1.2.3)
    maven-build-cache-config.xml (defaults)
  upstream/
    pom.xml
    src/main/java/example/Upstream.java
  downstream/
    pom.xml        (depends on upstream)
    src/main/java/example/Downstream.java
```

Steps:

```
mvn clean compile   # BUILD SUCCESS
mvn clean compile   # BUILD FAILURE
```

Happy to attach a zip or post a gist with the full project.

yeroc:
Fixes #492

## Description

`restoreProjectArtifacts` only wires up `project.getArtifact()` to a restored artifact when the cached file is a regular jar. For compile-only cached builds the primary cached artifact is the `target/classes` directory, which falls through that condition unset. In a multi-module reactor that leaves downstream modules unable to resolve the upstream as a reactor artifact on a subsequent `mvn clean compile`, and they fall through to remote repositories with `Could not find artifact ...:jar:...-SNAPSHOT`.

For the directory case this PR sets the file on the existing project artifact instead of replacing it with a `RestoredArtifact`. `RestoredArtifact.getFile` routes through `restoreToDiskConsumer` on every call, which would later break package-phase plugins (e.g. `jar:jar`) that update the file via `setFile`.

Added `CompileOnlyReactorRestoreTest`, which reuses the existing `issue-393-compile-restore` reactor project to exercise two consecutive compile-only invocations. The test fails on master and passes with this change.

## Checklist

### DIFF
--- a/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
+++ b/src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
@@ -481,6 +481,13 @@ public class CacheControllerImpl implements CacheController {
                 if (!project.hasLifecyclePhase("package")) {
                     project.addLifecyclePhase("package");
                 }
+            } else if (restoredProjectArtifact != null) {
+                // Directory artifact (target/classes from compile-only build).
+                // Point the existing project artifact at the restored directory so reactor dependency
+                // resolution can locate this module's compiled output. Don't replace the artifact:
+                // RestoredArtifact.getFile() always re-routes through restoreToDiskConsumer, which would
+                // break later phases (e.g. jar:jar) that update the file via setFile().
+                project.getArtifact().setFile(restoredProjectArtifact.getFile());
             }
             restoredAttachedArtifacts.forEach(project::addAttachedArtifact);
             restorationReport.setSuccess(true);
--- a/src/test/java/org/apache/maven/buildcache/its/CompileOnlyReactorRestoreTest.java
+++ b/src/test/java/org/apache/maven/buildcache/its/CompileOnlyReactorRestoreTest.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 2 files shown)


yeroc:
I considered renaming the existing `issue-393-compile-restore` reactor project to something more generic since it covers a fairly common and important case with multi-module reactor and internal dependencies that isn't covered elsewhere as far as I can tell.  Thoughts on renaming that?

yeroc:
@sparsick looking at the failure for build # 2 it looked like a strange intermittent issue pulling down transitive dependencies from Maven Central.  The 3rd build was cancelled.  As far as I can tell the failure doesn't appear to be related to the changes I made? Thoughts?

yeroc:
@olamy any thoughts on this PR?

PR #502, opened 2026-07-10
Drop Wagon; is history

cstamas:
Drop Wagon, is historical relic. If needed, users can re-add it as dependency.

Also, some cleanup in properties and dependencies.


Bukama:
Wagon is not archived, so if you declare it here as that, you might also want to start a vote to archive it?

cstamas:
> Wagon is not archived, so if you declare it here as that, you might also want to start a vote to archive it?

I am not talking about "archiving" wagon, but stating the fact that wagon is relic, was created as _transport abstraction for Maven2_ and should not be used in newly created stuff, and it is unused here as well, so just remove it from classpath.

PR #509, opened 2026-07-18
Reverse Maven versions

cstamas:
Reverse Maven versions (build against stable, test with stable + others), and fix the ITs as they never run against wanted Maven version. Also, let's do first Resolver 1 vs Resolver 2 (as in Maven 3.9.x vs Maven 3.10.x), as currently available Maven 4-rc-5 has so many changes happened since (ie JDK transport), that we should wait for 4-rc-6. And this PR solves the Resolver 2 anyway.

Fixed several issues:
* added GH actions to cover Maven 3.9, 3.10. Maven 4 still has inherent issues and cannot made to work with BCE, so dropping it for now. Focus on Resolver 2 first, and we will deal with Maven 4 later.
* no need for "outer" Maven matrix, we do it inside (w/ profile), the build itself can run with single (stable) maven version, everything else is just wasteful
* added Resolver 1 vs Resolver 2 changes (properties, exception handling)
* Verifier 1.8.0 quirk: uses `--debug` for "verbose output" that is repurposed in Maven 4 for "remote debugging", causing Maven 4 to stall and wait for remote debugger attach.
* Fixed Issue #511 as it was wrongly passing
* and finally, the extension is not tested on Java 8 (nor was), as build is Java 11+ and this is the problem of single module builds that IMO should never be done. Solution: refactor build into multiple modules, see https://maveniverse.eu/blog/2024/09/07/very-simple-ci-setup/

Contains fix for #511

### DIFF
--- a/src/site/markdown/remote-cache.md
+++ b/src/site/markdown/remote-cache.md
@@ -143,7 +143,7 @@ from Maven 3.9.1 and 4.0.0-alpha5, support of webdav has been removed per defaul
 You need to use the following extra configuration:
 
 ```bash
--Daether.connector.http.supportWebDav=true
+-Daether.transport.http.supportWebDav=true
 OR
 -Dmaven.resolver.transport=wagon
 ```
--- a/src/site/markdown/remote-cache.md
+++ b/src/site/markdown/remote-cache.md
@@ -143,8 +143,8 @@ from Maven 3.9.1 and 4.0.0-alpha5, support of webdav has been removed per defaul
 You need to use the following extra configuration:
 
 ```bash
--Daether.transport.http.supportWebDav # mvn 3.10+
--Daether.connector.http.supportWebDav # mvn 3.9
+-Daether.transport.http.supportWebDav=true # mvn 3.10+
+-Daether.connector.http.supportWebDav=true # mvn 3.9
 OR
 -Dmaven.resolver.transport=wagon
 ```
--- a/src/site/markdown/remote-cache.md
+++ b/src/site/markdown/remote-cache.md
@@ -143,7 +143,8 @@ from Maven 3.9.1 and 4.0.0-alpha5, support of webdav has been removed per defaul
 You need to use the following extra configuration:
 
 ```bash
--Daether.transport.http.supportWebDav=true
+-Daether.transport.http.supportWebDav # mvn 3.10+
+-Daether.connector.http.supportWebDav # mvn 3.9
 OR
 -Dmaven.resolver.transport=wagon
 ```
--- a/src/test/java/org/apache/maven/buildcache/its/Issue463Test.java
+++ b/src/test/java/org/apache/maven/buildcache/its/Issue463Test.java
@@ -134,7 +134,8 @@ class Issue463Test {
         verifier.setAutoclean(false);
         verifier.addCliOption("--settings=" + settings);
         // Allow Aether to create WebDAV collection (directory) hierarchy via MKCOL
-        verifier.setSystemProperty("aether.transport.http.supportWebDav", "true");
+        verifier.setSystemProperty("aether.transport.http.supportWebDav", "true"); // Maven 3.10+
+        verifier.setSystemProperty("aether.connector.http.supportWebDav", "true"); // Maven 3.9
 
         // ── Step 1: Runner A builds with skipTests=true, saves incomplete entry to remote ────────
         verifier.setSystemProperty(CACHE_LOCATION_PROPERTY_NAME, localCacheA.toString());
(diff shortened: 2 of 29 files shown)


cstamas:
And now that tests runs against different Maven versions:
* 3.9.16 is green, as it was
* 3.10.0 fails as apache transport has DAV disabled, as documented; and more, just found out
* 4.0.0-rc-5 ?

cstamas:
With latest changes 3.9 and 3.10 should be green, 4 is still lost.

cstamas:
Fun fact: Verifier does not support Maven 4, so Maven 4 is not workable as long we don't migrate over to
https://github.com/apache/maven-verifier/blob/master/MIGRATION.md

cstamas:
My suggestion: let's pause on Maven 4 support, so will disable it in this PR, and move with 3.9 and 3.10 support and testing. Once we migrate over to maven-executor, we can bring Maven 4 in as well.

src/site/markdown/remote-cache.md
```diff
@@ -143,7 +143,8 @@ from Maven 3.9.1 and 4.0.0-alpha5, support of webdav has been removed per defaul
 You need to use the following extra configuration:
 
 ```bash
--Daether.connector.http.supportWebDav=true
+-Daether.transport.http.supportWebDav # mvn 3.10+
+-Daether.connector.http.supportWebDav # mvn 3.9
```
gnodet: **[Nit]** The previous version had `=true` (i.e. `-Daether.connector.http.supportWebDav=true`). While Maven CLI treats bare `-DpropName` as setting value `"true"`, users configuring this in POM `<properties>` or `.mvn/maven.config` contexts need `=true` explicitly (since `Boolean.parseBoolean("")` returns `false`). Adding `=true` back improves copy-paste safety:

```suggestion
-Daether.transport.http.supportWebDav=true # mvn 3.10+
-Daether.connector.http.supportWebDav=true # mvn 3.9
```

src/test/java/org/apache/maven/buildcache/its/Issue463Test.java
```diff
@@ -134,7 +134,8 @@ void runnerBShouldHitRemoteCacheAfterRunnerAUploadsCompleteEntry() throws Verifi
         verifier.setAutoclean(false);
         verifier.addCliOption("--settings=" + settings);
         // Allow Aether to create WebDAV collection (directory) hierarchy via MKCOL
-        verifier.setSystemProperty("aether.connector.http.supportWebDav", "true");
+        verifier.setSystemProperty("aether.transport.http.supportWebDav", "true"); // Maven 3.10+
+        verifier.setSystemProperty("aether.connector.http.supportWebDav", "true"); // Maven 3.9
```
gnodet, line 138: **[Suggestion]** The dual-property pattern here is correct and consistently applied across `BaselineDiffTest` and `SaveFinalRemoteTest`. However, `RemoteCacheDavTest.java` (lines 156, 173) still only sets the Resolver 1 property (`aether.connector.http.supportWebDav`). Consider also updating `RemoteCacheDavTest` with the Resolver 2 property for consistency — the current mixed approach (explicit here, implicit via `maven.config` there) creates fragility if someone later restructures the test project config.

gnodet (commented):
## Review Summary

This PR reverses the Maven version strategy — from building against Maven 4.0.0-alpha-8 (unstable) with a Maven 3 test profile, to building against Maven 3.9.16 (stable) with dedicated profiles for testing against 3.10.x and 4.0.x. The approach is architecturally sound: compile against the stable API surface, test across multiple runtimes.

The individual fixes are well-considered:
- `--debug` → `-X` in ITs (Maven 4 repurposed `--debug` for remote JPDA debugging) ✅
- Resolver 2 `HttpTransporterException` support in `RemoteCacheRepositoryImpl` ✅
- Dual WebDAV properties for Maven 3.9 / 3.10+ compatibility ✅
- Build/test version separation (`mavenVersion` for compile deps, `maven.version` for test distribution) ✅
- Fix for #511 (`projectVersioning` config in p10-reactor-partial) ✅

### Scanner Coverage

| Scanner | Status | Findings |
|---------|--------|----------|
| semgrep | ✅ Ran | 0 |
| PMD | ✅ Ran | 10 `GuardLogStatement` — all pre-existing |
| gitleaks | ✅ Ran | No secrets |
| ast-grep | ✅ Ran | 0 |

All review questions resolved with only minor suggestions. Nice work! 👏

---
_This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

cstamas:
Latest change re-enabled Maven 4 and while 3.9 and 3.10 are green, 4 has failures.

slawekjaranowski (approved):

PR #519, opened 2026-07-30
Add includeHidden option to <include> to opt hidden files back into cache input scanning

### LINKED ISSUE #518: Add includeHidden option to <include> to opt hidden files back into cache input scanning
Files and directories whose name starts with a dot are always treated as hidden and unconditionally skipped when scanning `<include>` paths for cache input files, with no way to override this.

This is too broad for project layouts where meaningful, versioned content is stored in dotfile-named files by convention (e.g. Maven Filevault/JCR content packages, where a resource's definition typically lives in a file literally named `.content.xml`). Such content changes are currently invisible to the build cache.

Proposal: add an `includeHidden` attribute (default `false`) to the `<include>` element, so a specific include path can opt back into scanning hidden entries without changing default behavior elsewhere.

I have a working patch (with unit + integration tests, full test suite passing) and will open a PR referencing this issue.

royteeuwen:
Closes #518

## Summary
Files/directories whose name starts with a dot are always treated as hidden and unconditionally skipped when scanning `<include>` paths for cache input files, with no way to override this. This is too broad for project layouts where meaningful, versioned content is stored in dotfile-named files by convention (e.g. Maven Filevault/JCR content packages, where a resource's definition typically lives in a file literally named `.content.xml`), making such content changes invisible to the cache.

## Change
Adds an `includeHidden` attribute (default `false`, fully backward compatible) to the `<include>` element, scoped per-include, so a specific include path can opt back into scanning hidden entries without affecting default behavior anywhere else.

## Testing
- Extended `MavenProjectInputTest` with two new unit tests (default still skips hidden files; `includeHidden="true"` collects them).
- Extended the existing `IncludeExcludeTest` end-to-end integration test fixture with a hidden-file scenario.
- Full test suite: `mvn clean install` (88/88 unit tests), `mvn verify -Prun-its` (333/333 IT tests) - all green, no regressions.


src/main/java/org/apache/maven/buildcache/checksum/MavenProjectInput.java
```diff
@@ -428,15 +434,31 @@ private Path convertToAbsolutePath(Path path) {
      */
     private void startWalk(
             Path candidate, String glob, boolean recursive, List<Path> collectedFiles, Set<WalkKey> visitedDirs) {
+        startWalk(candidate, glob, recursive, false, collectedFiles, visitedDirs);
```
olamy: no need to keep backward compat for private methods

src/main/java/org/apache/maven/buildcache/checksum/WalkKey.java
```diff
@@ -28,11 +28,17 @@ public class WalkKey {
     private final Path normalized;
     private final String glob;
     private final boolean recursive;
+    private final boolean includeHidden;
 
     public WalkKey(Path normalized, String glob, boolean recursive) {
+        this(normalized, glob, recursive, false);
```
olamy, line 38: mark this constructor `deprecated`

src/main/mdo/build-cache-config.mdo
```diff
@@ -951,6 +951,15 @@ under the License.
                     <type>String</type>
                     <description>Files in a directory are filtered by matching their names against this glob.</description>
                 </field>
+                <field xml.attribute="true">
```
olamy, line 954: we usually increase model version in case of model changes (see pom.xml)

olamy (changes requested):

src/main/java/org/apache/maven/buildcache/checksum/MavenProjectInput.java
royteeuwen: Good point — removed the overload and updated the call sites to pass `includeHidden` explicitly.

src/main/java/org/apache/maven/buildcache/checksum/WalkKey.java
royteeuwen: Done — added `@Deprecated` plus a javadoc `@deprecated` tag pointing at the 4-arg constructor. Nothing in the codebase uses the 3-arg one anymore, so it is there purely for compatibility.

src/main/mdo/build-cache-config.mdo
royteeuwen: Bumped the modello model version from `1.3.0` to `1.4.0` in `pom.xml`, and updated the namespace/schemaLocation in `src/site/resources/maven-build-cache-config.xml` to match — following the same pattern as c915399 (the last model change). I left the existing test project configs on their current namespaces, as that commit did too.

royteeuwen:
Thanks for the review @olamy — all three points addressed in fd8c5c5.

Verified locally: `mvn clean verify -Prun-its` → 88 unit tests and 333 ITs green (the 4 + 1 skips are pre-existing).

PR #532, opened 2026-09-11
Add configurable remote build cache retention strategies

royteeuwen:
## Summary

Adds opt-in remote build-cache retention for shared repositories, with configurable strategy providers and Nexus Raw support.

## Changes

- Adds remote retention configuration separate from local retention.
- Adds configurable maximum remote builds and cleanup grace period.
- Adds explicit retention strategy selection by name, required whenever cleanup is enabled.
- Adds Maven-native Plexus/Sisu retention strategy providers.
- Adds Nexus Raw asset discovery, pagination, metadata validation, and asset-ID deletion.
- Adds generic HTTP/WebDAV directory-listing retention with safe limitations.
- Adds public `RemoteCacheRetentionStrategyProvider`, `RemoteCacheRetentionStrategy`, and `RemoteCacheHttpClient` APIs.
- Adds cleanup for stale `build-cache-report.xml` assets.
- Adds best-effort concurrency protections without distributed locking.
- Documents configuration, SPI usage, authentication, concurrency behavior, and repository limitations.

## Configuration

```xml
<remote enabled="true"
        saveToRemote="true"
        cleanupEnabled="true"
        retentionStrategy="nexus">
    <url>https://nexus.example/repository/build-cache/</url>
    <maxBuildsCached>5</maxBuildsCached>
    <cleanupGracePeriodSeconds>300</cleanupGracePeriodSeconds>
</remote>
```

Existing configurations remain compatible because cleanup is disabled by default. Once `cleanupEnabled="true"` is set, `retentionStrategy` must name a registered provider (`nexus`, `directory-listing`, or `unsupported`) — configuration fails clearly if it is missing or unknown. There is no URL-based strategy guessing.

## Notes

Nexus Raw paths are virtual asset paths rather than real directories. The implementation deletes assets using the Nexus Assets REST API, but cannot delete empty virtual browse folders independently.

Concurrent builders are supported on a best-effort basis. The implementation does not introduce distributed locking.


olamy:
Hi,
Thanks a lot for your contribution. 
But to be honest, I do not think such server configurations should be done on the client side (as every single user can have different values, so this can cause conflicts and eventually delete the whole cache).
I have some local work on a sort of cache server with such a configuration, but everything is centralised on the server side.

royteeuwen:
@olamy thats indeed why i would assume that as a developer you use the local cache, but configure the remote cache and set it as disabled. In the CI pipeline you enable it with a parameter you pass on build, which would be feasable with this implementation. For the develop branch i'd write to the remote cache and for the PR branches I'd read from it.

But of course I'm all ears to alternatives, if you say you have another sort of implementation that achieves the same thing and is contributable? On github I could use the actions/cache and still configure it as local cache, but where I want to use this, it's Jenkins, where I dont have this infra available to cache locally :(

olamy:
> @olamy thats indeed why i would assume that as a developer you use the local cache, but configure the remote cache and set it as disabled. In the CI pipeline you enable it with a parameter you pass on build, which would be feasable with this implementation. For the develop branch i'd write to the remote cache and for the PR branches I'd read from it.
> 
> But of course I'm all ears to alternatives, if you say you have another sort of implementation that achieves the same thing and is contributable? On github I could use the actions/cache and still configure it as local cache, but where I want to use this, it's Jenkins, where I dont have this infra available to cache locally :(

I have been started this https://github.com/jenkinsci/maven-cache-plugin but TBH I didn't really look at improve this seriously. Maybe you can help?

Otherwise my work on a sort of cache server is really different.

royteeuwen:
Ah sadly enough I can't install jenkins plugins myself, that's managed on organisation level and I'm not sure if I could get this in. I will check

If i would pull out the nexus implementation to my own jar and just keep the SPI / interface, would it then be acceptable as contribution, or is it definitely a no-go? What are the downsides of the way i described?

- Local on developer station uses the XML-based configuration
- CI pipelines use -D arguments to enable remote caching (with ideally also the purging that I'm contributing)

PR #534, opened 2026-09-15
Include transitive SNAPSHOT inputs in project checksums

### LINKED ISSUE #407: Locally built transitive snapshot artifacts don't cause checksum change of a dependent module
### Affected version

1.2.1

### Bug description

Our setup might be a bit unique.
We have 2 maven multi module projects one libs and one apps (they share a parent multi module but it's irrelevant).

We compile the libs first and then the apps. All libs versions are snapshot versions.

When a change is made to one of the libs module, If the an app module does not directly reference said lib module its checksum is unchanged, as the effective pom is unchanged and looking the the buildinfo.xml for the app module it only includes the direct snapshot dependencies and not all transitive snapshots dependencies.

Not sure if there is a contributing factor but we also set the jar plugin outputTimestamp for a reproducible build, I can go into details why it's necessary if needed.

Thanks!

goutamadwant:
Fixes #407.

For `app -> bridge -> leaf-SNAPSHOT`, changing `leaf` can leave the reproducible `bridge` JAR byte-identical. The application checksum currently includes only declared mutable dependencies, so a clean application build can restore stale compiled constants from cache.

Collect Maven's mediated dependency graph before cache lookup and hash additional external SNAPSHOT artifacts. Use project checksums for transitive reactor inputs. If graph resolution is unavailable, skip cache restoration and saving, then let normal Maven execution enforce dependency requirements.

The integration tests cover SNAPSHOT and release bridges, unchanged-input cache hits, exclusions, dependency management, transitive reactor inputs, and offline `validate`. Both bridge cases fail on unchanged master (`8bec537`); their fixed cached outputs match uncached rebuilds. Graph collection happens before lookup; only additional external SNAPSHOT artifacts are resolved.

Validation on JDK 17:

- `mvn -Prun-its verify javadoc:jar`: passed with Maven 3.9.16. The run reported 97 unit tests and 354 integration tests, with zero failures or errors. Four schema-dependent unit tests and one macOS permission test were skipped.
- `mvn -Prun-its,maven3.10 -Dtest='MavenProjectInput*' -Dit.test=TransitiveSnapshotTest verify`: passed with Maven 3.10.0-rc-1 for the integration builds.

  This change uses integration tests against Maven's real dependency resolver.


src/main/java/org/apache/maven/buildcache/checksum/MavenProjectInput.java
```diff
@@ -665,7 +676,59 @@ private static boolean isReadable(Path entry) throws IOException {
...
     private SortedMap<String, String> getMutableDependencies() throws IOException {
-        return getMutableDependenciesHashes("", project.getDependencies());
+        SortedMap<String, String> hashes = getMutableDependenciesHashes("", project.getDependencies());
+        if (project.getDependencies().isEmpty()) {
+            return hashes;
+        }
+
+        // Collect through Maven so dependency management, exclusions and scope mediation match the build.
+        // Resolve only additional external snapshots; reactor dependencies use project checksums.
+        DefaultDependencyResolutionRequest request =
```
olamy, line 686: Well, now we are forcing full transitive graph collection + snapshot resolution during cache-key calculation, which could be avoided when you have cache hits,  for cache hits where we were skipping dependency resolution. 
On a large project, this will add unexpected potential latency

src/main/java/org/apache/maven/buildcache/CacheControllerImpl.java
```diff
@@ -199,7 +200,14 @@ public CacheResult findCachedBuild(
 
         String projectName = getVersionlessProjectKey(project);
 
-        ProjectsInputInfo inputInfo = projectInputCalculator.calculateInput(project);
+        final ProjectsInputInfo inputInfo;
+        try {
+            inputInfo = projectInputCalculator.calculateInput(project);
+        } catch (DependencyGraphResolutionException e) {
+            LOGGER.warn("Skipping build cache for {}: {}", projectName, e.getMessage());
```
olamy, line 207: Well, if any remote repositories check for snapshots, we will now skip a module that has not been restored or saved. Caching is becoming subject to remote repo availability; this sounds fragile to me.

src/main/java/org/apache/maven/buildcache/checksum/MavenProjectInput.java
```diff
@@ -665,7 +676,59 @@ private static boolean isReadable(Path entry) throws IOException {
...
+                new DefaultDependencyResolutionRequest(project, session.getRepositorySession());
+        request.setResolutionFilter((node, parents) -> {
```
olamy, line 688: there should be filtering here. for main source we do not need to check `test` scope dependencies. Any change in test dependencies will invalidate the cache for src/main

PR #539, opened 2026-09-23
Replace javax.annotation-api with jakarta.annotation-api 1.3.5

slachiewicz:
## Problem

`javax.annotation:javax.annotation-api` has `net.java:jvnet-parent` as its parent, which declares the long-shut-down `https://maven.java.net` repositories. Dependency-POM repositories are not used for artifact resolution, but they do leak into the remote repository set used for **metadata** lookups — so every build pays a failing HTTPS round-trip against the dead host. Maven 4's plugin-prefix discovery probes it as well:

```
[WARNING] Could not transfer metadata /.meta/prefixes.txt from/to jvnet-nexus-releases
(https://maven.java.net/content/repositories/releases/): (certificate_expired) PKIX path validation failed
```

## Change

`javax.annotation:javax.annotation-api:1.3.2` → `jakarta.annotation:jakarta.annotation-api:1.3.5`.

The jakarta artifact is the EE4J continuation of the same JSR-250 API, published under the **same `javax.annotation` package names** (jakarta package rename only happened in 2.x), so this is a drop-in replacement — no source changes needed. The project only uses `@PostConstruct` and `@Priority` from it; the `javax.annotation.Nonnull`/`Nullable` imports resolve from `com.google.code.findbugs:jsr305` (unchanged), not from this artifact. The EE4J parent declares no active resolution repositories.

## Verification

- `mvn test-compile` — clean
- `mvn test -Dtest=LifecyclePhasesHelperTest,BuildInfoTest,BuildCacheMojosExecutionStrategyTest` — 35/35 pass (covers the `@PostConstruct` and `@Priority` usages)


slachiewicz:
Pushed a second commit: removed unused dependencies found by `dependency:analyze` + source inspection (`maven-resolver-util`, `maven-resolver-transport-http`, the three slf4j test bridges), and moved `maven-verifier` (only used by `its` tests via `org.apache.maven.it.Verifier`) from compile to test scope. `jmh-generator-annprocess` is kept deliberately (annotation processing for `PerfTest`). Verified: `test-compile` clean, 62/62 targeted unit tests pass, `dependency:analyze` now only reports the JMH false positive, `spotless:apply` clean.

pom.xml
```diff
@@ -242,24 +230,6 @@ under the License.
       <version>${slf4jVersion}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
olamy, line 245: not related to the title of the PR and breaking the build by the way

olamy (changes requested):

